PR #1, opened 2019-06-18
Update XMaterial.java

WI5E:
Code improvements


WI5E:
- mat could potentially return null because XMaterial::requestOldXMaterial has the possibility of returning null
- When calling Enum#values it will create a new array each time. If you're going to be executing the call multiple times, cache it once and use it.

PR #3, opened 2019-10-02
Fix Terracotta

Drc-DEV:
Basically XMaterial prioritized recognizing GLAZED terracotta over normal terracotta. This meant that having STAINED_CLAY:1 as an input in pre-flattening versions for XMaterial.matchXMaterial, it would have returned glazed terracottas instead of normal ones. XMaterial just returns the first match from the list, so I just swapped glazed and normal terracotta's order in the list.


CryptoMorin:
Sorry I didn't see this at all.
This should already be fixed in v3

PR #5, opened 2019-12-12
Optimizations & Code Improvements

WI5E:
- Refactored ImmutableMap to ImmutableEnumMap for performance improvements
- Used stream method reference over direct method invocation
- Replaced string utility implementation with one provided by a Spigot lib
- Removed condition redundancy


CryptoMorin:
Nice one. Thanks.

PR #6, opened 2019-12-26
Fix all dyes in pre 1.13

PatoTheBest:
The old name in the Material enum was INK_SACK, new name is INK_SAC

https://hub.spigotmc.org/stash/projects/SPIGOT/repos/bukkit/browse/src/main/java/org/bukkit/Material.java?until=5ad2ecf658276b8aae336b0e5bfb894a5151b578&untilPath=src%2Fmain%2Fjava%2Forg%2Fbukkit%2FMaterial.java#422


CryptoMorin:
Thanks. The reason I used `INK_SAC` is because I was trying to make the requestOldMaterial Method find the legacy material from the XMaterial enum with the newest name and then parse the old name from that enum. But it wasn't really efficient so I forgot to change it.

PatoTheBest:
No worries, thank you for this awesome resource!

PR #12, opened 2020-01-24
Maven repository

portlek:


src/main/java/org/cryptomorin/xseries/ActionBar.java
```diff
@@ -65,7 +67,7 @@
      * @see #sendActionBarWhile(Player, String, Callable)
      * @see #sendActionBarWhile(Player, Callable, Callable)
      */
-    private static final JavaPlugin PLUGIN = ;
+    private static JavaPlugin PLUGIN;
```
portlek: Improve this field with `setInstance(org.bukkit.plugin.Plugin)` method.

portlek: Done with tricky way that's `Bukkit.getPluginManager().getPlugins()[0]`.

CryptoMorin:
Oh yeah, I saw this. I already mentioned that I'm going to make a maven repo myself.
And issues with ActionBar has been fixed (in an upcoming version)

Also, it's not recommended to just use another plugin for schedulers. They're important factors for server performance analysis and other tasks.

PR #15, opened 2020-01-27
Issue template update.

portlek:


CryptoMorin:
I really think it's good the way it is right now. [Feature] and {Bug] titles are just redundant since it'll be labeled anyway, adding it to the title is useless.

PR #20, opened 2020-04-19
Update README.md

portlek:


CryptoMorin:
How does the **${version}** work tho? Is there anywhere I can read about it?

portlek:
> How does the **${version}** work tho? Is there anywhere I can read about it?

It won't work,It's just a placeholder that users should replace with the maven-central case.

CryptoMorin:
I see, can you change it to **version** then? It kinda looks like a bug when you set it as ${version}

portlek:
If I change it, we shall change it every version changing. So this is a problem actually. We may forget to change the version number. Users should know the latest version unless we change it in the readme.

CryptoMorin:
No no. I mean change it to "version". Literally the word "version".

portlek:
alright

PR #21, opened 2020-04-19
Add XChatColor

TheSilentPro:
A utility class for more ChatColor methods

### DIFF
--- a/src/main/java/com/cryptomorin/xseries/XChatColor.java
+++ b/src/main/java/com/cryptomorin/xseries/XChatColor.java
@@ -1,0 +1,118 @@
+package com.cryptomorin.xseries;
+
+import org.bukkit.ChatColor;
+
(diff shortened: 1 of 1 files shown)


CryptoMorin:
I'm not sure how this is supposed to benefit any plugin. It just looks really random.

TheSilentPro:
> I'm not sure how this is supposed to benefit any plugin. It just looks really random.

Not really benefits, but rather narrowing down the code they'd have to use.
I know this might be kind of random from what the XSeries is supposed to be intended for use.

If it doesn't fit the project, I completely understand.

CryptoMorin:
I really appreciate that you're trying to improve the project, however this looks like something that 1% of the developers might use. If anyone wanted to make something similar, they'd mostly make their utility to suit their plugin.

PR #26, opened 2020-05-08
[XSound] Fixes usage for Player#stopSound on Bukkit v1.9 and lower

CodingAir:
The method does not exist for Bukkit v1.9 and lower.

### DIFF
--- a/src/main/java/com/cryptomorin/xseries/XSound.java
+++ b/src/main/java/com/cryptomorin/xseries/XSound.java
@@ -28,10 +28,7 @@ import com.google.common.cache.CacheBuilder;
 import org.apache.commons.lang.StringUtils;
 import org.apache.commons.lang.Validate;
 import org.apache.commons.lang.WordUtils;
-import org.bukkit.Instrument;
-import org.bukkit.Location;
-import org.bukkit.Note;
-import org.bukkit.Sound;
+import org.bukkit.*;
 import org.bukkit.entity.Entity;
 import org.bukkit.entity.Player;
 import org.bukkit.plugin.java.JavaPlugin;
@@ -1055,8 +1052,7 @@ public enum XSound {
 
         return CompletableFuture.runAsync(() -> {
             for (XSound music : musics) {
-                Sound sound = music.parseSound();
-                if (sound != null) player.stopSound(sound);
+                music.stopSound(player);
             }
         });
     }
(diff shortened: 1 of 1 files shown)


portlek:
I think it should throw an exception about the support versions. Devs. would like to know what's the problem.

CodingAir:
> I think it should throw an exception about the support versions. Devs. would like to know what's the problem.

Yeah, but using an extra try catch or if statement to prevent this error would be messy as well. And if a developer would take a look at the function, (s)he would see, that there is no function for it on 1.9 and lower.

src/main/java/com/cryptomorin/xseries/XSound.java
```diff
@@ -1210,6 +1207,9 @@ public void run() {
     public void stopSound(@Nonnull Player player) {
         Objects.requireNonNull(player, "Cannot stop playing sound from null player");
 
+        int version = Integer.parseInt(Bukkit.getBukkitVersion().split("-", -1)[0].split("\\.")[1]);
```
CryptoMorin, line 1210: Extremely inefficient. Check if a 1.10 sound is available using Enums.getIfPresent method and save it as a constant. If there are no new sounds added in 1.10, just get a material with Material.getMaterial()

CryptoMorin:
Also, just reminding anyone who's reading this. I accept these as temporary solutions since the main file has changed a lot. I'm waiting for 1.16

src/main/java/com/cryptomorin/xseries/XSound.java
CodingAir: What about this? It would parse the version number into a byte once. 

```
private static byte version = Byte.parseByte(Bukkit.getBukkitVersion().split("-", -1)[0].split("\\.")[1]);

    public void stopSound(@Nonnull Player player) {
        Objects.requireNonNull(player, "Cannot stop playing sound from null player");

        if(version < 10) return; //Player#stopSound is not available for v1.9 and lower

        Sound sound = this.parseSound();
        if (sound != null) player.stopSound(sound);
    }

```

CryptoMorin: I have no idea why you're parsing it as a byte and not an integer lol. Why not just use it as a simple boolean constant? We're not using it anywhere else. Also, make it final too.

PR #34, opened 2020-06-27
Build fix.

portlek:
Build fix for newbie users of maven.


CryptoMorin:
Okay wait, what do you mean "The library is from http://nathaantfm.alwaysdata.net/explorer/?package=com.mojang&name=authlib"? I never heard of "nathaantfm.alwaysdata.net" It's not an official link. If someone wants to import the profiles thing, they just have to either import Bukkit or Spigot Jars.

portlek:
Forget the authlib or something like that, why we don't use nms repository where is here `https://repo.codemc.org/repository/nms/` 
I think, we should add this repository for travis or other CI, we can add more test for every classes and we can make the library more stable. so i don't want to explore bugs of the library i just want to make it more safety for every project.

CryptoMorin:
Same problem, codemc is not an official repository. Yes, it works, but it's not official.

CryptoMorin:
Actually after a second thought, considering how much easier it'll be using this... I might just switch to codemc.
I'll think about it.

PR #49, opened 2020-07-18
Fixed fern data id

einTosti:
Fern data id should be 2 instead of 1


CryptoMorin:
Hmm, then which LONG_GRASS has a data value of 1?

einTosti:
According to [*this*](https://minecraft-ids.grahamedgecombe.com) site normal grass has the value of 1, which was also the item that appeared when I previously used XMaterial.FERN

Screenshot 2020-07-18 at 14 39 26

CryptoMorin:
I see. I think I need to add a legacy name and data value for the [grass](https://github.com/CryptoMorin/XSeries/blob/e7c5e4da233fd934c899ddf0beb6840beb3b41ab/src/main/java/com/cryptomorin/xseries/XMaterial.java#L496) enum.

PR #53, opened 2020-07-24
[XItemStack] Fixed parseColor method

robertlit:
Blue color is now gotten from the correct index


CryptoMorin:
Oops, I don't know why I didn't notice that.

PR #55, opened 2020-07-29
Clear ActionBars and added Lambdas

TheProgramSrc:
+ Added Lambda in line 192
+ Added Clear ActionBar method (From line 195 to 214)


CryptoMorin:
Please do not use lambda for `Bukkit.getOnlinePlayers()`  Although it works for 1.8.8 servers, it'll still break for versions lower than 1.8.8 Yes, people should be using the latest version, but we can avoid this easily.

TheProgramSrc:
Ok, I'll remove them

PR #72, opened 2020-10-07
Switch to Jetbrains annotations

iGabyTM:


pom.xml
```diff
@@ -51,6 +51,11 @@
     </repositories>
 
     <dependencies>
+        <dependency>
+            <groupId>org.jetbrains</groupId>
+            <artifactId>annotations</artifactId>
+            <version>20.1.0</version>
+        </dependency>
```
portlek, line 58: The score should be provided. The annotation classes are not effective within the classpath. It's for Intellij IDEA.

portlek (changes requested):

pom.xml
knightzmc: bruh what does that mean

iGabyTM: What do you mean? 🤔

portlek: scope*

portlek: I'm talking specifically about the annotations from Jetbrains, not other annotation libraries.

CryptoMorin:
You provided no description why we should switch to IntelliJ annotations.

portlek:
> You did not describe why we should switch to JetBrains annotations.

It's adding inspections for Intellij IDEA. This method or field can be null or not in compile time, not runtime.

CryptoMorin:
The thing is, for some reasons IntelliJ highlights the warnings under the actual method parameter which shows you the usages with nulls. It doesn't highlight the warning on the usage itself. So it doesn't matter which nullability annotations you use. You can't see the warning if it's in the library. I don't even see a warning for it when I compile.
Or is there a new option added in IntelliJ that fixes this?

portlek:
If you add your annotations in your dependencies(not from libraries) you can see.

CryptoMorin:
I added JetBrains annotations to both XSeries and my other project libraries via Maven and it's still not showing the error. If it's only about showing which method returns null or not documentation-wise, it works for all annotations.

iGabyTM:
As far as I know javax was removed from the jdk and Jetbrains' annotations is what most people use. Regarding the warnings, I see them on any project which has a library that uses the annotations.

CryptoMorin:
Alright, my IntelliJ setting were old. I fixed them and now it's showing. I used both in XSeries and without needing to add them to my main project's classpath, it showed the warning just fine. It works with any nullability annotations.

PR #77, opened 2020-10-28
[XItemStack] Added support for enchanted books

SirLeezus:
Added support for enchanted books specifically, tested on Spigot 1.8-1.16.

### DIFF
--- a/src/main/java/com/cryptomorin/xseries/XItemStack.java
+++ b/src/main/java/com/cryptomorin/xseries/XItemStack.java
@@ -111,15 +111,6 @@ public final class XItemStack {
             config.set(entry, enchant.getValue());
         }
 
-        // Enchanted Books
-        if (meta instanceof EnchantmentStorageMeta) {
-            EnchantmentStorageMeta book = (EnchantmentStorageMeta) meta;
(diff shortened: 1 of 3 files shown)


CryptoMorin:
What about **ItemMeta#getEnchants()**? That doesn't work on enchanted books? If not, then instead of **stored-enchants** path use **enchants** path which is used for the normal enchantments.

SirLeezus:
Sweet, that works for me! **ItemMeta#getEnchants()** also does work but it isn't considered the same as a normal enchanted book, they take stored enchants and doing it this way allows you to compare them with the normal enchanted books in the game.  In my case I needed it for a GUI menu and I wanted to allow players to sell normal enchanted books but if I applied normal enchants to a enchanted book they wouldn't compare.

SirLeezus:
I also just attempted switch the path to **stored-enchants** to **enchants** but it would load normal enchants on top of stored enchants doing that sadly.

SirLeezus:
Hey! I'll open a new pull request for this, I ended up making a new fork, couldn't get my account to login correctly.

CryptoMorin:
Okay, I'll leave this open. You can close it any time.

PR #79, opened 2020-10-30
[XItemStack] Enchanted Book Stored Enchants & Shulker Box Support

SirLeezus:
Added Enchanted Book Stored Enchants Support!

Let me know if you would like anything changed, switching the path to enchants didn't work correctly for me and for whatever reason you want to load enchanted books with stored enchants and normal enchants it can be done this way.

I also added shulker box item support in one of the commits if you wanted to take a look.


CryptoMorin:
Looks pretty good. Thank you very much.

PR #80, opened 2020-10-30
[XItemStack] Spawner Serialize Support & Fixed Fireworks

SirLeezus:
Small one but I needed it.

Also, in later versions like 1.8 on deserialize **spawner.update(true);** does not work on itemstacks. You'll get an error saying something along the lines of you can only update a Block. I fixed it so it would work for 1.8-1.16 with the following:
**bsm.setBlockState(spawner); 
item.setItemMeta(meta);**

Just thought I would let you know, I'm aware other people use this for different things so might not work correctly for them in some cases so I didn't commit it.

I also added a fix in for fireworks, I had to use two array lists because for some reason the colors were not getting serialized correctly. This is the firework I used for testing:
```
firework:
          '0':
            type: BALL_LARGE
            flicker: true
            trail: false
            base-colors:
            - 59, 81, 26
            - 65, 205, 52
            fade-colors:
            - 59, 81, 26
          '1':
            type: CREEPER
            flicker: false
            trail: true
            base-colors:
            - 65, 205, 52
            fade-colors:
            - 65, 205, 52
```

### DIFF
--- a/src/main/java/com/cryptomorin/xseries/XBiome.java
+++ b/src/main/java/com/cryptomorin/xseries/XBiome.java
@@ -130,6 +130,7 @@ public enum XBiome {
      * @since 1.0.0
      */
     public static final List<XBiome> VALUES = Collections.unmodifiableList(Arrays.asList(values()));
+    private static final boolean supports = XMaterial.supports(16);
     @Nullable
     private final Biome biome;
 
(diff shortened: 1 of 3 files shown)


CryptoMorin:
I'm not sure why it's not showing the conflicts, but I changed XItemStack a little after merging yours. Can you use that one please?

SirLeezus:
I think I updated it correctly, if not feel free to fix it as you wish, my only goal is to make you aware.

I did test the changes you made to shulker boxes, I only found one issue with serialize, the line:
**ConfigurationSection shulkerSection = config.getConfigurationSection("shulker");**
will always be null because it's never set in the config

PR #82, opened 2020-10-30
Biomes fix

IridiumLLC:
In the new Nether update introduced in 1.16, you can have different biomes on different y-axies. This pull request will fix this


IridiumLLC:
Ok, ima redo this cuz im an idiot and didnt use Branches

PR #83, opened 2020-10-30
[XBiome] Fixed issues for setting biomes on 1.16+

IridiumLLC:
In the Nether update, they introduced the ability to have different biomes depending on the Y axies. This will fix setting the biome on 1.16+ versions

### DIFF
--- a/src/main/java/com/cryptomorin/xseries/XBiome.java
+++ b/src/main/java/com/cryptomorin/xseries/XBiome.java
@@ -130,6 +130,7 @@ public enum XBiome {
      * @since 1.0.0
      */
     public static final List<XBiome> VALUES = Collections.unmodifiableList(Arrays.asList(values()));
+    private static final boolean supports = XMaterial.supports(16);
     @Nullable
     private final Biome biome;
 
@@ -167,7 +168,8 @@ public enum XBiome {
         for (int i = 0; i < len; i++) {
             char ch = name.charAt(i);
 
-            if (!appendUnderline && count != 0 && (ch == '-' || ch == ' ' || ch == '_') && chs[count] != '_') appendUnderline = true;
+            if (!appendUnderline && count != 0 && (ch == '-' || ch == ' ' || ch == '_') && chs[count] != '_')
+                appendUnderline = true;
             else {
                 if ((ch >= 'A' && ch <= 'Z') || (ch >= 'a' && ch <= 'z')) {
                     if (appendUnderline) {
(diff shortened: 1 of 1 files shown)


src/main/java/com/cryptomorin/xseries/XBiome.java
```diff
@@ -167,7 +168,8 @@ private static String format(@Nonnull String name) {
         for (int i = 0; i < len; i++) {
             char ch = name.charAt(i);
 
-            if (!appendUnderline && count != 0 && (ch == '-' || ch == ' ' || ch == '_') && chs[count] != '_') appendUnderline = true;
+            if (!appendUnderline && count != 0 && (ch == '-' || ch == ' ' || ch == '_') && chs[count] != '_')
+                appendUnderline = true;
```
CryptoMorin, line 172: Please change this line back.

src/main/java/com/cryptomorin/xseries/XBiome.java
```diff
@@ -240,9 +242,14 @@ public Biome getBiome() {
         // Apparently setBiome is thread-safe.
         return CompletableFuture.runAsync(() -> {
             for (int x = 0; x < 16; x++) {
-                for (int z = 0; z < 16; z++) {
-                    Block block = chunk.getBlock(x, 0, z);
-                    if (block.getBiome() != biome) block.setBiome(biome);
+                // y loop for 1.16+ support (vertical biomes).
+                // As of now increasing it by 4 seems to work.
+                // This should be the minimal size of the vertical biomes.
```
CryptoMorin, line 247: > This should be the minimal size of the vertical biomes.

Can you also reference the source that why it should be specifically 4?

CryptoMorin: It'd be even better if there's a source available from either Spigot or Minecraft that explains the biome system in 1.16

CryptoMorin (changes requested):

src/main/java/com/cryptomorin/xseries/XBiome.java
IridiumLLC: > > This should be the minimal size of the vertical biomes.
> 
> Can you also reference the source that why it should be specifically 4?

Trial and error, 4 seemed to be the biggest number before issues started occuring :shrug:

CryptoMorin: Ah ok, I see. Thanks

PR #84, opened 2020-10-30
[XBiome] Added info on what world the biome is designed for

IridiumLLC:
As Minecraft is adding different biomes for different worlds, I thought it would be useful to get what world the biome is designed for.


CryptoMorin:
Umm, you removed the 1.16 biome fix in your latest commit.

IridiumLLC:
ffs, ima just redo the pull req, gime like 5 mins

PR #88, opened 2020-11-22
[XEnchantment] QUICK_CHARGE Enchant Fix

SirLeezus:
Simple one but I had to fix it for one of my plugins so I thought I would let you know.


CryptoMorin:
Thank you :)

PR #90, opened 2020-11-25
WIP: [XMaterial] Add MAP as a Duplicated Material

thenilsdev:
I am aware that support for everything under 1.13 is not guaranteed, but I would like to provide this change as it will certainly help some developers.


CryptoMorin:
But what is this fixing exactly? What didn't work before (please show with a code example) that it works with MAP added to this list?

thenilsdev:
In my understanding https://github.com/CryptoMorin/XSeries/blob/97d3dcb378d51cb4d2251434e4288ad2b1ae20f7/src/main/java/com/cryptomorin/xseries/XMaterial.java#L450 is a duplicate of https://github.com/CryptoMorin/XSeries/blob/97d3dcb378d51cb4d2251434e4288ad2b1ae20f7/src/main/java/com/cryptomorin/xseries/XMaterial.java#L695.

Same for example on https://github.com/CryptoMorin/XSeries/blob/97d3dcb378d51cb4d2251434e4288ad2b1ae20f7/src/main/java/com/cryptomorin/xseries/XMaterial.java#L696 and https://github.com/CryptoMorin/XSeries/blob/97d3dcb378d51cb4d2251434e4288ad2b1ae20f7/src/main/java/com/cryptomorin/xseries/XMaterial.java#L698.

Without this change an empty map resolves into `FILLED_MAP` and not `MAP`.

Not tested on higher version and as I saw in the comment, I'll look further into it, to fix it.

CryptoMorin:
Any updates on this?

CryptoMorin:
This has been closed due to inactivity.

PR #92, opened 2020-12-27
Add XSeriesTypeAdapterFactory for Gson usage

ReflxctionDev:
I use XSeries extensively in my projects, and especially, side-by-side with Gson. By default, Gson serializes enums with `Enum#toString()`, and deserializes them back with `Enum#valueOf(String)`. While this works fine in most cases, it may not be optimal for XSeries's enums, which have to do legacy checks and so on.

For this, I wrote up this small TypeAdapterFactory to use the XSeries methods `XSomething#matchXSomething`, and return the result from the Optional, or throw an IllegalArgumentException with easy to identify messages, such as "Invalid biome: mars". 

Essentially, I wrote this for my own usage, but I'm sure it may be beneficial for other users of this amazing library. Let me know if anything needs modification.

### DIFF
--- a/src/main/java/com/cryptomorin/xseries/XSeriesTypeAdapterFactory.java
+++ b/src/main/java/com/cryptomorin/xseries/XSeriesTypeAdapterFactory.java
@@ -1,0 +1,85 @@
+package com.cryptomorin.xseries;
+
+import com.google.common.collect.ImmutableMap;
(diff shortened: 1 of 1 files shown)


CryptoMorin:
> which have to do legacy checks and so on.
They don't have to do anything legacy checks. Just use valueOf() without even worrying about exceptions.

PR #95, opened 2021-01-08
[XMaterial] - Spruce Wood & Dark Oak having the same data

OhSry:
Spruce wood kept on getting confused with DARK_OAK due to the data being the same in spruce_log and dark_oak_log/wood, so I removed "LOG" from dark_oak_log/wood.


CryptoMorin:
Thanks!

PR #97, opened 2021-01-08
[XMaterial] Get an XMaterial from a Block

OhSry:
This could be useful instead of people having to get the type and the data then convert it into an ItemStack just to get an XMaterial out of it.


CryptoMorin:
There's a reason behind XBlock. **block.getData()** doesn't work since it's different for rotational, directional and ... blocks.

PR #106, opened 2021-03-02
[XItemStack] Major Changes - Map, Book & SpawnEgg Support

DeadSilenceIV:
Hello i added the BookMeta and the MapMeta into the project, also the SkullMeta serialization and deserialization was fixed for legacy versions. I also fixed some unsupported methods for legacy versions.

Issue #91, opened 2020-12-08
[XMaterial] - Filled Map #0 resolves into MAP instead of FILLED_MAP in legacy versions


LOOHP:
**Description**
Filled Maps in 1.8 resolves into MAP (which is for empty maps) instead of FILLED_MAP

In version 1.8.9:
Opens a map in the game
XMaterial material = XMaterial.matchXMaterial(playerItemStackInHand);
material -> XMaterial.MAP (which should be XMaterial.FILLED_MAP)

**Version**
This bug occurs on 1.8.9 but it works fine on 1.16.4

CryptoMorin:
This only happens to **Map #0* which really can't be fixed since both this and EMPTY_MAP have a data value of 0 and there isn't a way to detect whether it's a filled map or not.

The only way is to check if the item meta is an instance of [MapMeta](https://hub.spigotmc.org/javadocs/bukkit/org/bukkit/inventory/meta/MapMeta.html) which is not something that XMaterial is supposed to do. Adds an extra check only to support a single material.

LOOHP:
Ah alright, then I'll do the check myself. Thank you.

PR #107, opened 2021-03-03
[XMaterial] Fixed FILLED_MAP Matching in Legacy Versions

DeadSilenceIV:


DeadSilenceIV:
Any updates on this topic?

CryptoMorin:
I'm just going to leave a description for future references.

Original Issue: #106 (Also related to #91)
> This change was made because in legacy versions the filled map is saved as "MAP" and at the moment of deserializing it will take the form of an empty map instead of a filled map (Ik, it's kind of confusing but i've already tested it and it is working as it should for legacy and newer versions), You can try yourself with the current version and will see the issue i'm refering.

In legacy versions (1.12 and below) Minecraft assigns a new data value (durability) every time a map is created (FILLED_MAP). The increment starts from **0**

---

Tested using **1.8.9 and 1.12.2 PlayerInteractEvent** comparing how it behaves differently. 
First by clicking on a **MAP** material and then clicking on a **FILLED_MAP**:
```java
ItemStack item = player.getItemInHand();
XMaterial mat = XMaterial.matchXMaterial(item);
print(mat.name() + " - " + mat.parseMaterial() + " - " + mat.parseItem() 
        + " | &9" + item.getType() + " (" + item.getDurability() + ")");
```
The blue part is supposed to be the same obviously.
The green text compares the material using this new system and the red text compares with the current system:

Screenshot 2021-03-17 233316
---
But when used on a fresh world as the first map:

Screenshot 2021-03-18 000418

CryptoMorin:
This needs to be optimized further, but I'll take care of that.

PR #108, opened 2021-03-05
Adds XVersion to make methods that are relevant to Bukkit server version independent.

portlek:


CryptoMorin:
Let's not please... I want the classes to be standalone. It's not doing any harm to the performance. Design is debatable, but like I said, I want them to be independent.

It might be questionable that why XBiome and XEntity should depend on XMaterial, but the core of this class is XMaterial. Other utilities of this class that depend on the version are rarely used without using XMaterial too.

So, if you really want to change it, it's better to write independent version checks (they don't necessarily need to implement the full system in XMaterial) for those utilities.

PR #110, opened 2021-03-05
Adds delegation for Bukkit's configuration section.

portlek:
That allows developers to use serialize/deserialize methods with their own configuration section implementations.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -63,6 +63,12 @@
 
     <dependencies>
         <dependency>
+            <groupId>org.projectlombok</groupId>
+            <artifactId>lombok</artifactId>
+            <version>1.18.18</version>
+            <scope>provided</scope>
+        </dependency>
+        <dependency>
             <groupId>org.spigotmc</groupId>
             <artifactId>spigot</artifactId>
             <version>1.16.4-R0.1-SNAPSHOT</version>
--- a/src/main/java/com/cryptomorin/xseries/XEntity.java
+++ b/src/main/java/com/cryptomorin/xseries/XEntity.java
@@ -21,6 +21,8 @@
  */
 package com.cryptomorin.xseries;
 
+import com.cryptomorin.xseries.section.BukkitConfigurationSection;
+import com.cryptomorin.xseries.section.XConfigurationSection;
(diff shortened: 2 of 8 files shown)


portlek:
For example, this is my own configuration section interface;
image

CryptoMorin:
I'm planning to use a map instead like you suggested a long time ago. This is just too overkill.

portlek:
> I'm planning to use a map instead like you suggested a long time ago. This is just too overkill.

Yea I suggested but I think this is also a simple way to implement so I did it.

portlek:
Also, about map as a parameter, maybe this may be a little bit buggy or slow because if you want to add backward compatibility for the deserialize/serialize methods, you must make a method that converts the sections in the section map into a real map, It's something like that;
```java
@NotNull
default Map<String, Object> getMapValues(final boolean deep) {
  return this.getValues(deep).entrySet().stream()
    .map(entry -> {
      final var key = entry.getKey();
      final var value = entry.getValue();
      if (value instanceof ConfigurationSection) {
        return new AbstractMap.SimpleEntry<>(key, ((ConfigurationSection) value).getMapValues(deep));
      }
      return new AbstractMap.SimpleEntry<>(key, value);
    })
    .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue));
}
```

DeadSilenceIV:
Hello portlek, i would like to know what would be the benefits of using the **BukkitConfigurationSection** class instead of the normal **ConfigurationSection** class (Just for curiosity).

portlek:
> Hello portlek, i would like to know what would be the benefits of using the **BukkitConfigurationSection** class instead of the normal **ConfigurationSection** class (Just for curiosity).

if someone uses their own configuration section library like Simple-YAML or others, you can't use the serialization method, so XSeries have to use their own configuration section interface on the methods then BukkitConfigurationSection must delegate/wrap the original Bukkit configuration section. That's it. It looks like a stupid idea but it works :D

CryptoMorin:
Right now, the problem is I want to keep these classes as independent as possible.
I'll have to think about it.

PR #116, opened 2021-05-20
[ParticleDisplay] Add force support

cricri211:
I already added this in my XParticle usage, I think everyone should be able to use force, with false by default.

Feel free to not accept this PR since force particles can be resource-intensive for clients with small GPU configurations !

### DIFF
--- a/src/main/java/com/cryptomorin/xseries/particles/ParticleDisplay.java
+++ b/src/main/java/com/cryptomorin/xseries/particles/ParticleDisplay.java
@@ -54,7 +54,7 @@ import java.util.concurrent.Callable;
  * <code>[r, g, b, size]</code>
  *
  * @author Crypto Morin
- * @version 5.0.0
+ * @version 5.0.1
  * @see XParticle
  */
 public class ParticleDisplay implements Cloneable {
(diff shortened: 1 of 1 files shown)


portlek (approved):

cricri211:
I have literally no ideas of what does the error mean :')

```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.8.1:compile (default-compile) on project XSeries: Compilation failure
[ERROR] Bad service configuration file, or exception thrown while constructing Processor object: javax.annotation.processing.Processor: Provider org.eclipse.sisu.space.SisuIndexAPT6 not found
```

Does it come from the changed here or something ?

cricri211:
Oh, I didn't know the link was dynamic, sorry for the spam.

It seems the fail also happens without this change.

Im-Fran:
Hey @cricri211 I had the same issue with maven but I solved it by adding the following configuration to the maven compiler plugin:
```xml
<configuration>
    <proc>none</proc> <!-- This is what I added -->
    <source>1.8</source>
    <target>1.8</target>
</configuration>
```

I don't know why but it just stopped working, because I never created a custom annotation processor
[Original Answer from StackOverflow](http://stackoverflow.com/questions/38926255/ddg#54281656)

CryptoMorin:
Yeah, I'm not sure why that option is needed in the latest version of IntelliJ. It was working just fine before.

cricri211:
Close this PR and reopen another one with proper checks and no reverts !

Issue #114, opened 2021-04-16
[ParticleDisplay] - Add support for colored ITEM_CRACK, BLOCK_CRACK, BLOCK_DUST and FALLING_DUST


Drc-DEV:
**Description**
why does XParticle#filledCircle with a ParticleDisplay#colored object throw java.lang.IllegalArgumentException: data should be class java.lang.Void got class org.bukkit.Particle$DustOptions ?
I basically want to create filledCircle with a colored MOB_SPELL particle (insted of the default REDSTONE particle)
My code:
```
ParticleDisplay dust = ParticleDisplay.colored(location,r,g,b,1);
dust.particle = XParticle.getParticle("MOB_SPELL");
XParticle.filledCircle(radius,rate,rrate,dust);
```
I also tried using a particledisplay from ParticleDisplay.simple, adding the color via #withColor, but the result is the same exception.

any help?

cricri211:
The spell particle behavior is really basic and frustating. Colored particle with XParticle works only with `REDSTONE` (which is `dust`, in vanilla) and Bukkit API cannot deal with colored spell like it does with dust.

TL;DR : The main issue is actually vanilla, if you want an example : https://gaming.stackexchange.com/questions/352449/change-the-colour-of-a-particle. You can use colored spell particle with their offset.

#
### Let's go back to the basic behavior of particles with /particle command

1. A classic particle has a dx dy dz parameters, the offset of where it can spawn around the given location, then a speed and a count, the number or particle to be displayed. In short, this is : `/particle <particle> x y z dx dy dz speed count`.


2. Some particles can be directional, like flame, nautilus, end rod, totem, ... it means that when the count is 0, that is we want to spawn "0" particles, it will display one but with a velocity. The velocity is given by dx dy dz. The command becomes : `/particle <particle> x y z velocityX velocityY velocityZ speed 0`. For example, a single flame particle that goes east will have a dx = 1, dy = 0, dz = 0. Higher the velocity is, more speed the particle has (in addition to the speed parameter). Note that a directional particle spawns at the exact given location without any offset, because the offset becomes the velocity.


3. Other specific particles directly modify the particle command syntax. It is the case of : 
- dust (REDSTONE) : `/particle dust red green blue size x y z dx dy dz speed count normal` (Sadly, dust particles cannot be directional, a count of 0 result to a single motionless particle)
- item (ITEM_CRACK) : `/particle item <item name> x y z ...`
[...]

PR #120, opened 2021-05-25
[ParticleDisplay] Force support and item/block datas

cricri211:
There are two added features :
- Force support, like I did in #116.
- Support for BLOCK_CRACK, BLOCK_DUST, FALLING_DUST and ITEM_CRACK particles, customizable with data field and methods withBlock and withItem (#114). **In theory, the feature supports version 1.12 but I have not tested it.**
- It "fixed" issue #114 since it also allows particles like SPELL_MOB or SPELL_MOB_AMBIENT to use colors in the same way as REDSTONE does. When SPELL_MOB or 1.12 REDSTONE has count=0 (directionnal), the offset becomes the color and the speed becomes other thing, for REDSTONE it's size but for SPELL_MOB it's a mix of offset, speed and saturation color. In 1.13, REDSTONE changed to get other parameters, but not SPELL_MOB.

-> Tested in 1.16.5 with some recent Paper builds in java 15, but not in other versions ! (In 2 or 3 weeks, if this pr hasn't been merged yet, I'll test it on other versions)

Commit description : 
>• BLOCK_CRACK, BLOCK_DUST, FALLING_DUST and ITEM_CRACK are now supported in the data field.
>• Data field becomes private to prevent users from freely changing its value. A getter is added and methods withColor, withBlock and withItem modify this value.
>• Method withEntity shouldn't get a null entity to avoid a NullPointerException everytime the entity is null.
>• Deserialisation with edit method is adapted to blocks and item supports. Not tested in 1.12 or less.
>• Spawn particle method is also adapted. It can spawn MOB_EFFECT particle with color with in same way as it does with 1.12 REDSTONE particle (color becomes the offset) if count=0.

### DIFF
--- a/src/main/java/com/cryptomorin/xseries/particles/ParticleDisplay.java
+++ b/src/main/java/com/cryptomorin/xseries/particles/ParticleDisplay.java
@@ -24,10 +24,14 @@ package com.cryptomorin.xseries.particles;
 import org.apache.commons.lang.StringUtils;
 import org.apache.commons.lang.math.NumberUtils;
 import org.bukkit.Location;
+import org.bukkit.Material;
 import org.bukkit.Particle;
+import org.bukkit.block.data.BlockData;
 import org.bukkit.configuration.ConfigurationSection;
 import org.bukkit.entity.Entity;
 import org.bukkit.entity.Player;
+import org.bukkit.inventory.ItemStack;
+import org.bukkit.material.MaterialData;
 import org.bukkit.util.Vector;
 
 import javax.annotation.Nonnull;
@@ -54,7 +58,7 @@ import java.util.concurrent.Callable;
  * <code>[r, g, b, size]</code>
  *
  * @author Crypto Morin
- * @version 5.0.1
+ * @version 5.1.0
  * @see XParticle
  */
 public class ParticleDisplay implements Cloneable {
@@ -72,9 +76,9 @@ public class ParticleDisplay implements Cloneable {
     public double extra;
     @Nullable
     public Vector rotation;
-    @Nullable
-    public Object data;
     public boolean force;
+    @Nullable
+    private Object data;
 
     /**
      * Make a new instance of particle display.
@@ -121,8 +125,21 @@ public class ParticleDisplay implements Cloneable {
     }
 
     /**
+     * Get the data object. Currently, it can be instance of float[] with [R, G, B, size],
+     * or instance of {@link BlockData}, {@link MaterialData} for legacy usage or {@link ItemStack}
+     *
+     * @return the data object.
+     * @since 5.1.0
+     */
+    @Nullable
+    public Object getData() {
+        return data;
+    }
+
+    /**
      * Builds a simple ParticleDisplay object with cross-version
      * compatible {@link org.bukkit.Particle.DustOptions} properties.
+     * Only REDSTONE particle type can be colored like this.
      *
      * @param location the location of the display.
      * @param size     the size of the dust.
@@ -141,6 +158,7 @@ public class ParticleDisplay implements Cloneable {
     /**
      * Builds a simple ParticleDisplay object with cross-version
      * compatible {@link org.bukkit.Particle.DustOptions} properties.
+     * Only REDSTONE particle type can be colored like this.
      *
      * @param location the location of the display.
      * @param color    the color of the particle.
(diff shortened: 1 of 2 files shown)


src/main/java/com/cryptomorin/xseries/particles/ParticleDisplay.java
```diff
@@ -268,8 +303,25 @@ public static ParticleDisplay edit(@Nonnull ParticleDisplay display, @Nonnull Co
...
+        } else if (item != null) {
+            Material material = Material.getMaterial(item);
+            if (material != null && material.isItem()) {
+                display.data = new ItemStack(material, 1);
+            }
+        } else if (materialdata != null) {
+            Material material = Material.getMaterial(materialdata);
+            if (material != null && material.isBlock()) {
+                display.data = material.getData();
+            }
```
CryptoMorin, line 320: What's the point of having 3 different options that all accept a material? Do ItemStack, BlockData and MaterialData make a difference in the visual representation of the particle?

CryptoMorin:
I really appreciate that you put so much effort into this. The force support seems pretty straightforward, but I'm not sure if it's worth supporting a few colored particles because the code looks Really polluted right now. I might need some time to test this myself to understand what's going on before deciding on this.

src/main/java/com/cryptomorin/xseries/particles/ParticleDisplay.java
cricri211: - ItemStack only matches ITEM_CRACK particle (named "item" in vanilla).
- BlockData matches with BLOCK_CRACK (when a block is broken), BLOCK_DUST (When you run on a block) and FALLING_DUST (like falling sand particles below a flying sand block).
- MaterialData is for legacy usage, it's the magic id that modify the data. BlockData replaces MaterialData since 1.13 with a proper implementation, gread parameters (allowing us to use "stone_stairs[half=bottom,facing=north]" instead of "stone_stairs 7"). So, in 1.12 BLOCK_CRACK, BLOCK_DUST and FALLING_DUST are using MaterialData, but in 1.13 it's deprecated LEGACY_BLOCK_CRACK, LEGACY_BLOCK_DUST and LEGACY_FALLING_DUST.

In vanilla (or with Bukkit API), if one of this particles is used but no data is set, display fails. If one of this particles is used but a wrong data type is set, an error occurs.

I saw you support 1.12 (e.g. ISFLAT for DustOption) so I kept support for it even if users are in 1.13 or above. The best way is probaly to prevent users to set a data that doesn't matches the data type of the particle, but it needs to check the particle many times (on deserialization, when the particle changes, so accessors for particle field must be needed, and so on...).

My idea was to let users set a good type, knowing what he does, but if he uses a wrong type, it displays the particle without too many issues, with other parameters.

It's a kind of draft, it may be possible to set a Material in data field (no MaterialData, no BlockData, no ItemStack) and create the good type (MaterialData, BLockData or ItemStack) in spawns method depending on the particle, like for REDSTONE particle which uses a float[] instead of DustOption directly. But overload spawn method is possibly unsafe for performances, as it's the most frequently called method in XParticle.

cricri211:
Yeah I understand. Actually, I linked the SPELL_MOB and AMBIENT_SPELL_MOB particles to a color identically to the 1.12 REDSTONE particle. Another possibility is to lock color for REDSTONE and let users modify their offset and speed to get the color they want with spell_mob particle. It's tricky to understand that without a great minecraft commands knowledge.
I think SPELL_MOB and AMBIENT_SPELL_MOB are the only particles to have this strange behavior without extra data, so leave them is possible.

I did not find any other solutions to support block_crack, block_dust, falling_dust and item_crack, except maybe lock more the accessibility (q.v. my reply to your review). Maybe data field can be instance of an interface, and every impleting classes represents a particle data type ? So in spawn method, call "data.getValue(particle)" to get an item stack if the given particle is ITEM_CRACK, or a DustOption if the particle is REDSTONE ?

src/main/java/com/cryptomorin/xseries/particles/ParticleDisplay.java
CryptoMorin: I see, that makes sense. Thanks for the explanation!

CryptoMorin:
I think I finally understand what's happening. Everything looks great. The 1.12 support just needs a little changing. I'm going to take care of that myself.
Thanks a lot again!

PR #124, opened 2021-06-15
Switch to Github Actions.

portlek:
there is no reason to switch but idk if you want just merge it.


CryptoMorin:
Yea, if there's no reason, let's not.

PR #125, opened 2021-06-17
since 1.17 world height can be higher than 256

PeachesMLG:
Still need to do something about y starting at 0 but idk how to fix that so ill leave that to somone else


CryptoMorin:
Why not just use [getMinHeight()](https://hub.spigotmc.org/javadocs/spigot/org/bukkit/World.html#getMinHeight())
Also, you should cache the entire condition part of the loop in a variable. It seems like **getMaxHeight()** is calculated in 1.17

CryptoMorin:
I'm just going to add it myself.

PR #130, opened 2021-07-23
Fix titles not working on 1.17+

SkillC0de:


CryptoMorin:
I see, thanks. Your formatting changed a few lines, but don't worry I'll fix that.

PR #133, opened 2021-08-01
Update XMaterial.java

UntouchedOdin0:
To me it makes more sense to get a stick instead of gravel as its what represents it the closest.


CryptoMorin:
Uhhh, what? Gravel? That is the [Minecraft version](https://github.com/CryptoMorin/XSeries/blob/0ddc7d5c4e511b11a05f5d418b5b0b71ce7b9fb8/src/main/java/com/cryptomorin/xseries/XMaterial.java#L1533) that the material was added in...

UntouchedOdin0:
I mean, you currently have it so if they are on a lower version to give them gravel instead of a trident right? I thought it'll be more logical to give them a stick as it represents a trident more?

CryptoMorin:
That thing was removed a long long time ago.

PR #134, opened 2021-08-09
[SkullUtils] Cache the reflections for the profile field getter

ReflxctionDev:
In `SkullUtils#getSkinValue(ItemMeta)`, for some reason the `Field` instance is fetched every time the method is invoked. I presume this is not intended. This PR fixes it.


CryptoMorin:
Those field names are really confusing now tho. I'll change it.

Issue #142, opened 2021-11-04
[XItemStack] - Add support for serialization/deserialization of ItemStack directly from a Map


SecretX33:
**Description**
I would like to add support for serialization and deserialization directly from a `Map<String, Object>`, since I use `XItemStack` to handle serialization in a bunch of projects, but I am always capped by the fact that I am forced to use `MemoryConfiguration` when I need to serialize items, and I can't pass a `Map<String, Object>` to be serialized back to an ItemStack (for example, a Map read from a config file that is not handled by Bukkit `FileConfiguration` system).

If you agree, I'll start working on, and eventually push a PR adding support for serialization and deserialization using `Map`s, and ofc I won't touch the current methods that accept `ConfigurationSection`.

PR #143, opened 2021-11-06
Added serialization support for Maps

### LINKED ISSUE #141: [XItemStack] - NullPointerException when deserializing item with material AIR
**Description**

`NullPointerException` is thrown when serializing an ItemStack that is made of `AIR`. That happens because `XItemStack#serialize` calls  `getItemMeta` on the item passed as a parameter and uses the result of it without checking if it is null before.

**Code (Kotlin)**
```kt
XItemStack.serialize(ItemStack(Material.AIR, 1), MemoryConfiguration())
```

**Stacktrace:** https://paste.helpch.at/sajoraruko.cs

**Versions**

Java: `16`
XSeries: `8.4.0`
Kotlin: `1.5.31`
Gradle: `7.2`
Shadow Plugin: `7.0.0`
Paper: `1.16.5 (788)`

SecretX33:
As mentioned [here](https://github.com/CryptoMorin/XSeries/issues/142), here's the PR introducing support for item serialization and deserialization to and from `Map`. The new code works just fine as far as I could test, both serialization methods (using `ConfigurationSection` and passing a `Map<String, Object>` directly) did produce the same output.

I got some NPEs along the process and did fix all of them in one commit, the exceptions I got can be seen below and validated as well. Feel free to modify anything, or ask for modifications.

## Features

- Added serialization and deserialization support for `Map<String, Object>`.

## Fixes

- Fixed #141 
- Fixed NPE when serializing items that had AttributeModifiers that weren't bound to a specific slot (so `slot` was `null`).
- Fixed NPE when deserializing items that had AttributeModifier, that was happening because instead of querying the specific config entry of that Attribute, the code was querying the map of attributes, which of course does not have any attribute of specific properties on its root.

### DIFF
--- a/src/main/java/com/cryptomorin/xseries/XItemStack.java
+++ b/src/main/java/com/cryptomorin/xseries/XItemStack.java
@@ -36,6 +36,8 @@ import org.bukkit.block.ShulkerBox;
 import org.bukkit.block.banner.Pattern;
 import org.bukkit.block.banner.PatternType;
 import org.bukkit.configuration.ConfigurationSection;
+import org.bukkit.configuration.MemoryConfiguration;
+import org.bukkit.configuration.MemorySection;
 import org.bukkit.enchantments.Enchantment;
 import org.bukkit.entity.Axolotl;
 import org.bukkit.entity.EntityType;
@@ -113,6 +115,8 @@ public final class XItemStack {
             config.set("damage", item.getDurability());
         }
 
+        if (meta == null || !item.hasItemMeta()) return;
+
         // Display Name & Lore
         if (meta.hasDisplayName()) config.set("name", meta.getDisplayName());
         if (meta.hasLore()) config.set("lore", meta.getLore()); // TODO Add a method to "untranslate" color codes.
@@ -150,14 +154,13 @@ public final class XItemStack {
                     config.set(path + "name", modifier.getName());
                     config.set(path + "amount", modifier.getAmount());
                     config.set(path + "operation", modifier.getOperation().name());
-                    config.set(path + "slot", modifier.getSlot().name());
+                    if (modifier.getSlot() != null) config.set(path + "slot", modifier.getSlot().name());
                 }
             }
         }
 
         if (meta instanceof BlockStateMeta) {
-            BlockStateMeta bsm = (BlockStateMeta) item.getItemMeta();
-            BlockState state = bsm.getBlockState();
+            BlockState state = ((BlockStateMeta) meta).getBlockState();
 
             if (supports(11) && state instanceof ShulkerBox) {
                 ShulkerBox box = (ShulkerBox) state;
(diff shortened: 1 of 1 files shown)


src/main/java/com/cryptomorin/xseries/XItemStack.java
```diff
@@ -113,6 +115,8 @@ public static void serialize(@Nonnull ItemStack item, @Nonnull ConfigurationSect
             config.set("damage", item.getDurability());
         }
 
+        if (meta == null || !item.hasItemMeta()) return;
```
CryptoMorin, line 118: I'm pretty sure `ItemStack#hasItemMeta()` will also take care of `item == null`
Plus, this should be before `item.getItemMeta()` is actually invoked.

You should also take care of that `Damageable` meta. Just move `ItemStackMeta meta =...` after setting amount and `hasItemMeta` check before it.

src/main/java/com/cryptomorin/xseries/XItemStack.java
```diff
@@ -378,6 +394,8 @@ public static ItemStack deserialize(@Nonnull ConfigurationSection config) {
             if (damage > 0) item.setDurability((short) damage);
         }
 
+        if (meta == null) return item;
```
CryptoMorin, line 397: Same here, this should be done before `ItemMeta meta = ...` and move this after setting the amount.

src/main/java/com/cryptomorin/xseries/XItemStack.java
```diff
@@ -749,6 +770,52 @@ public static ItemStack deserialize(@Nonnull ConfigurationSection config) {
...
+     * @return a deserialized ItemStack.
+     */
+    @Nullable
+    public static ItemStack deserialize(@Nonnull Map<String, Object> serializedItem) {
+        Objects.requireNonNull(serializedItem, "serializedItem cannot be null.");
+        return deserialize(mapToConfigSection(serializedItem));
+    }
+
+    @Nonnull
+    private static ConfigurationSection mapToConfigSection(@Nonnull Map<?, ?> serializedItem) {
```
CryptoMorin, line 787: Is there any situation where `Map<String, Object>` is not available?

CryptoMorin (changes requested):

src/main/java/com/cryptomorin/xseries/XItemStack.java
portlek: or separating meta serialization and item stack serialization to 2 methods.

src/main/java/com/cryptomorin/xseries/XItemStack.java
SecretX33: Well, I don't know the internals of the deserialized map, maybe the deserialized map has `Integer` as keys at some point (because this method was designed to be called recursively, so the user might have some mapping from `Integer` to something at some nested level, or `int[]` as values, so I just went for the safe method, avoiding any `ClassCastException` that may arise there, plus it's not a big deal to use `Map<?, ?>` in that part of the code.

src/main/java/com/cryptomorin/xseries/XItemStack.java
SecretX33: But moving that check to after setting the amount (but before setting the durability) won't affect `1.8` potions (water or not) or something like that? I'm not exactly sure amount moving it to before setting the durability because of that.

CryptoMorin: Hmm, actually this line is wrong
> config.set("material", item.getType().name());

It's supposed to go thru XMaterial first. `XMaterial.matchXMaterial(item).name()` should fix it.
That'll take care of the data values and all that stuff because it's serialized with back with XMaterial.

SecretX33: I've modified that line. Is there anything else?

CryptoMorin (approved):

PR #150, opened 2021-12-11
Added XTags

RohanGoyalDev:
A possible replacement for  [Spigot's Tag system](https://hub.spigotmc.org/javadocs/bukkit/org/bukkit/Tag.html) that can be used for 1.8 and can have XSound, XBlock, XParticle genericised later on to ease in finding enum values that all carry a common property


RohanGoyalDev:
I'll add back generics later, didn't see he removed them

src/main/java/com/cryptomorin/xseries/XMaterial.java
```diff
@@ -2071,6 +2071,21 @@ public boolean isSupported() {
...
+    /**
+     * Returns a set of all supported materials out of the set provided
+     * <p>
+     *
+     * @param materials {@link Set} of the materials to be filtered
+     *
+     * @return {@link Set} of materials that contains only the supported materials out of the provided {@link Set}
+     */
+
+    public static Set<XMaterial> filterSupported(Set<XMaterial> materials) {
```
CryptoMorin, line 2083: Why is this useful?

src/test/XTagTest.java
```diff
@@ -0,0 +1,27 @@
...
+
+  @Test
+  void checkNonInheritTag() {
+    final XMaterial target = XMaterial.TUBE_CORAL;
+
+    assertTrue(XTag.CORAL_PLANTS.isTagged(target));
+  }
+
+  @Test
+  void checkInheritTag() {
```
CryptoMorin, line 20: You should add all tests to `Tests` class. I personally don't see the point of separating these.

src/main/java/com/cryptomorin/xseries/XTag.java
```diff
@@ -0,0 +1,2221 @@
+package com.cryptomorin.xseries;
+
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.Set;
+import org.checkerframework.checker.nullness.qual.NonNull;
+import org.checkerframework.checker.nullness.qual.Nullable;
+
+public final class XTag {
```
CryptoMorin, line 10: Wouldn't it make more sense to use an enum to avoid this mess?

CryptoMorin (changes requested):

CryptoMorin:
I'm personally not that familiar with tags, but have you looked at [XMaterialUtil](https://github.com/CryptoMorin/XSeries/blob/master/src/main/java/com/cryptomorin/xseries/XMaterialUtil.java)? Should we remove that and merge it with XTag?

> can have XSound, XBlock, XParticle genericised later on to ease in finding enum values that all carry a common property
Also, if this means changing those classes, discuss this before making any changes.

src/main/java/com/cryptomorin/xseries/XTag.java
CryptoMorin: Also, using static imports sounds reasonable at this point.

src/main/java/com/cryptomorin/xseries/XTag.java
```diff
@@ -0,0 +1,2221 @@
...
+        MUSHROOM_GROW_BLOCK.inheritFrom(NYLIUM);
+        MOSS_REPLACEABLE = new XTag();
+        MOSS_REPLACEABLE.inheritFrom(CAVE_VINES,
+                DIRT,
+                BASE_STONE_OVERWORLD);
+    }
+
+    private XTag(@NonNull XMaterial... values) {
+        this.values =
+            Collections.unmodifiableSet(new HashSet<>(Arrays.asList(values)));
```
CryptoMorin, line 2192: Use an EnumSet

RohanGoyalDev:
> I'm personally not that familiar with tags, but have you looked at [XMaterialUtil](https://github.com/CryptoMorin/XSeries/blob/master/src/main/java/com/cryptomorin/xseries/XMaterialUtil.java)? Should we remove that and merge it with XTag?
> 
> > can have XSound, XBlock, XParticle genericised later on to ease in finding enum values that all carry a common property
> 
> Also, if this means changing those classes, discuss this before making any changes.

Anything that is an enum can be used under XTag, so XBlock and XParticle can't rn but XPotion, XSound, XBiome, XEnchantment can without any changes in those enums

CryptoMorin:
I think you should really fix the enum issue before adding more stuff.

RohanGoyalDev:
> I think you should really fix the enum issue before adding more stuff.

Can you gimme like an example or class hierarchy? I thought of many ways to use an enum for this but all of them still create a mess

CryptoMorin:
Hmm, yeah you're right, after trying a few methods myself there's no way to make this cleaner while keeping the type safety or the cleanness from an API perspective.

RohanGoyalDev:
I will keep adding more Tags. If you or anyone else has a good idea on cleanup gimme an example and ill refactor the code accordingly if that works

CryptoMorin:
I think we should support tagged mappings too.
isTagged will still return a boolean without being affected, but we can add a new method of(key)
The only tradeoff is the performance and memory of `EnumMap` which will be extremely memory inefficient for a class like XMaterial because it has a lot of enums since `EnumMap` doesn't do Jumbo optimizations.

We can fix that if we make our own wrapper.
https://pastebin.com/BQEBypwb

Btw while testing this I noticed you did this:
```java
if (this.values.isEmpty()) newValues = EnumSet.copyOf((EnumSet<T>) this.values);
else newValues = EnumSet.copyOf(this.values);
```
There's no difference between these two lines, was this intentional?



We could also take Spigot's approach and use a switch statement in a simple static method instead of a map.

RohanGoyalDev:
> ```java
> if (this.values.isEmpty()) newValues = EnumSet.copyOf((EnumSet<T>) this.values);
> else newValues = EnumSet.copyOf(this.values);
> ```


Well those lines are intentional because of the illegal argument that `EnumSet.copyOf()` throws when the `Collection` passed in as a parameter is empty. And there are Tags like `ITEMS_FURNACE_MATERIALS` which are for some apparent reason left empty in spigot 1.18. To account for those and the tags that just inherit values from other tags and have no seperate value I had to made a seperate constructor because `T.class` is not a thing and `EnumSet.noneOf()` needs that, if i try to solve both of those issues then I cant make the `values` class variable an unmodifiable set because `Collections.unModifiableSet()` returns a set that cant be casted to `EnumSet`

RohanGoyalDev:
> > ```java
> > if (this.values.isEmpty()) newValues = EnumSet.copyOf((EnumSet<T>) this.values);
> > else newValues = EnumSet.copyOf(this.values);
> > ```
> 
> Well those lines are intentional because of the illegal argument exception that `EnumSet.copyOf()` throws when the `Collection` passed in as a parameter is empty UNLESS its another `EnumSet`. And there are Tags like `ITEMS_FURNACE_MATERIALS` which are for some apparent reason left empty in spigot 1.18. To account for those and the tags that just inherit values from other tags and have no seperate value I had to made a seperate constructor because `T.class` is not a thing and `EnumSet.noneOf()` needs that, if i try to solve both of those issues then I cant make the `values` class variable an unmodifiable set because `Collections.unModifiableSet()` returns a set that cant be casted to `EnumSet`

CryptoMorin:
If you're done with this, tell me to merge it.

RohanGoyalDev:
I think its done for now apart from the tagged mappings. I'll keep adding more tags which i see useful in more pull requests later on

PR #165, opened 2022-03-02
Fixed unsupported material with XMaterial#matchItemStack(ItemStack)

Gypopo:
**Description**

Using XMaterial#matchXMaterial(new ItemStack(Material.GREEN_DYE)) says its an unsupported material on **1.13.2**
**Stacktrace:** https://pastebin.com/ByEBTrGM

### DIFF
--- a/src/main/java/com/cryptomorin/xseries/XMaterial.java
+++ b/src/main/java/com/cryptomorin/xseries/XMaterial.java
@@ -1684,12 +1684,9 @@ public enum XMaterial {
         String material = item.getType().name();
         byte data = (byte) (Data.ISFLAT || item.getType().getMaxDurability() > 0 ? UNKNOWN_DATA_VALUE : item.getDurability());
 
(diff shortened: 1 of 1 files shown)


Gypopo:
Added a new commit to solve a error with XMaterial#matchXMaterial(POTION) returned a unknown material.

Gypopo:
I also removed the check for if its a monster_egg to check for its SpawnEggMeta since SpawnEggMeta doesn't exist in 1.8 and would also throw a error, you might add this back if its intended.

LOOHP:
Perhaps check whether SpawnEggMeta exists before checking SpawnEggMeta instead?

CryptoMorin:
Yeah, you can't just remove something entirely as a solution.

CryptoMorin:
Fixed in v8.7.0

Btw, SpawnEggMeta exists even in 1.8.8, Idk what version you tested this with.

PR #179, opened 2022-06-25
Fix material data for ACACIA & DARK_OAK. (#178)

### LINKED ISSUE #178: [XMaterial] - ACACIA_LOG & DARK_OAK_LOG data
**Description**

``ACACIA_LOG`` seems to be missing byte data (should be ``4``). This results in the material not matching correctly using the ``#requestOldXMaterial(String, byte)`` method.

``DARK_OAK_LOG`` and ``DARK_OAK_LEAVES`` have data defined, but with value ``1``. However from debugging it seems like the correct value should be ``5`` - like other block types form this type of material (``DARK_OAK_PLANKS, SLABS``).

**Version**
Tested on 1.12.2, 1.8.8
XSeries 8.8.0

Just a side note and proposal. I've started using ``#matchXMaterial(String)`` with the name composed of ``type:data`` because there's no API endpoint available to match material *and* data on legacy versions. I think it would be convenient to provide one in the lines of ``#matchXMaterial(Material, byte)``.

Wertik:
- correct material data for ACACIA & DARK_OAK
- fix mismatched legacy type for STRIPPED_DARK_OAK_LOG & WOOD

Resolves #178 

*All changes tested on 1.12.2*


CryptoMorin:
I believe `STRIPPED_DARK_OAK_LOG` is a new material in 1.13 right? In that case it shouldn't have a legacy name, I'll remove that.
Thanks for fixing the other materials.

PR #181, opened 2022-06-29
Fix music disc material IDs in legacy Minecraft versions

I2000C:
In legacy Minecraft versions, there are some music discs which ID doesn't match with their real ID.
I have changed them, so they match their actual ID.

image

In the image, the discs at the bottom are music discs obtained with the next fragment of code:

image

and the discs at the top are the discs in the correct order


CryptoMorin:
Thank you so much

PR #183, opened 2022-08-01
Replacing XMaterial.supports(13) with ISFLAT

MrMarL:
Why not?


CryptoMorin:
Sure, but don't make the same for the ones in XMaterial, they're intended to be like that.

PR #185, opened 2022-08-19
Added XMaterial#isAir which works the same as Material#isAir

DavidTheExplorer:


CryptoMorin:
I add enum matching methods in general to [XTag](https://github.com/CryptoMorin/XSeries/blob/master/src/main/java/com/cryptomorin/xseries/XTag.java) class instead.

PR #187, opened 2022-08-29
Fixed data values with some ACACIA and DARK_OAK materials

I2000C:
In legacy Minecraft versions, some ACACIA and DARK_OAK materials had incorrect data values.
I have changed them to their correct value.
The affected xmaterial list is:
- ACACIA_LEAVES (changed from 4 to 0)
- ACACIA_LOG (changed from 4 to 0)
- ACACIA_WOOD (changed from 4 to 0)
- DARK_OAK_LEAVES (changed from 5 to 1)
- DARK_OAK_LOG (changed from 5 to 1)
- DARK_OAK_WOOD (changed from 5 to 1)

In the image, the items at the left have been obtained using XMaterial and the items at the right are the correct items that correspond to first items.
acacia_dark_oak_error

Code used to obtain the XMaterial items of the example:
image


CryptoMorin:
Thanks a lot for the detailed explanation. 👍

PR #199, opened 2022-12-24
playAtLocation option for NoteBlockMusic & Cloneable Record

HSGamer:
Add an option to toggle either the whole world or only the player can hear the music

### DIFF
--- a/src/main/java/com/cryptomorin/xseries/XSound.java
+++ b/src/main/java/com/cryptomorin/xseries/XSound.java
@@ -1937,7 +1937,7 @@ public enum XSound {
      *
      * @since 3.0.0
      */
-    public static class Record {
+    public static class Record implements Cloneable {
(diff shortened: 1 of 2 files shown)


src/main/java/com/cryptomorin/xseries/XSound.java
```diff
@@ -1961,6 +1961,10 @@ public Record(@Nonnull XSound sound, @Nullable Player player, @Nullable Location
             this.playAtLocation = playAtLocation;
         }
 
+        public Record(Record other) {
+            this(other.sound, other.player, other.location, other.volume, other.pitch, other.playAtLocation);
```
CryptoMorin, line 1965: Why not implement `Cloneable` instead?

CryptoMorin:
Thanks!

PR #201, opened 2023-01-03
Added EnumUtils which removes the code duplication of creating a display name for enums.

DavidTheExplorer:
This PR initially started with statically importing all **Collectors**, but then I realized they were all involved in generating elegant display names for enums - so instead I added a class that removes all the code duplications.

### DIFF
--- a/src/main/java/com/cryptomorin/xseries/XEnchantment.java
+++ b/src/main/java/com/cryptomorin/xseries/XEnchantment.java
@@ -31,8 +31,10 @@ import org.bukkit.inventory.meta.EnchantmentStorageMeta;
 
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
+
+import static java.util.stream.Collectors.joining;
+
 import java.util.*;
-import java.util.stream.Collectors;
 
 /**
  * Enchantment support with multiple aliases.
@@ -308,7 +310,7 @@ public enum XEnchantment {
     public String toString() {
         return Arrays.stream(name().split("_"))
                 .map(t -> t.charAt(0) + t.substring(1).toLowerCase())
-                .collect(Collectors.joining(" "));
+                .collect(joining(" "));
     }
 
     /**
--- a/src/main/java/com/cryptomorin/xseries/XEnchantment.java
+++ b/src/main/java/com/cryptomorin/xseries/XEnchantment.java
@@ -21,6 +21,7 @@
  */
 package com.cryptomorin.xseries;
 
+import com.cryptomorin.xseries.utils.EnumUtils;
 import com.google.common.base.Enums;
 import org.bukkit.Material;
 import org.bukkit.NamespacedKey;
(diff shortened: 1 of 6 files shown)


CryptoMorin:
The class utilities are designed to work as a standalone class. Most of them shouldn't depend on other classes.

DavidTheExplorer:
@CryptoMorin At least merge the collectors commit, it helps with readability by a lot + the Collectors API was actually desgined to be statically imported.

CryptoMorin:
> @CryptoMorin At least merge the collectors commit, it helps with readability by a lot + the Collectors API was actually desgined to be statically imported.

Where exactly did you read that? Where's the source? It helps with the readability yes but "a lot" sounds like a huge exaggeration to actually change it.

Even OpenJDK uses it that way: [1](https://github.com/openjdk/jdk18/blob/6dae52f8e3993d529033147de8e34ad1e7d48c53/test/langtools/tools/javac/classfiles/attributes/annotations/ClassType.java#L33-L37) [2](https://github.com/openjdk/jdk18/blob/6dae52f8e3993d529033147de8e34ad1e7d48c53/src/jdk.javadoc/share/classes/jdk/javadoc/internal/doclets/toolkit/taglets/snippet/Attributes.java#L48-L50) [3](https://github.com/openjdk/jdk18/blob/6dae52f8e3993d529033147de8e34ad1e7d48c53/src/java.naming/share/classes/com/sun/jndi/ldap/LdapCtx.java#L2713-L2716)

Spigot: [1](https://hub.spigotmc.org/stash/projects/SPIGOT/repos/craftbukkit/browse/src/main/java/org/bukkit/craftbukkit/tag/CraftFluidTag.java#25) [2](https://hub.spigotmc.org/stash/projects/SPIGOT/repos/craftbukkit/browse/nms-patches/net/minecraft/util/worldupdate/WorldUpgrader.patch#24-25) [3](https://hub.spigotmc.org/stash/projects/SPIGOT/repos/craftbukkit/browse/src/main/java/org/bukkit/craftbukkit/entity/CraftPiglin.java#6,67,72)

Paper: [1](https://github.com/PaperMC/Paper/blob/1143b636634183fe9ee6d8cf9709afd4b3239de4/patches/api/0157-Add-Material-Tags.patch#L48) [2](https://github.com/PaperMC/Paper/blob/d5f1ffc2e6928a6728f91ab2ad2e020b105eab7f/patches/server/0755-Hide-unnecessary-itemmeta-from-clients.patch#L31-L32) [3](https://github.com/PaperMC/Paper/blob/d5f1ffc2e6928a6728f91ab2ad2e020b105eab7f/patches/server/0397-Deobfuscate-stacktraces-in-log-messages-crash-report.patch#L142-L143)

[Official documentation:](https://docs.oracle.com/javase/8/docs/technotes/guides/language/static-import.html)
> So when should you use static import? **Very sparingly**! Only use it when you'd otherwise be tempted to declare local copies of constants, or to abuse inheritance (the Constant Interface Antipattern). In other words, use it when you require frequent access to 
static members from one or two classes

That's what I'm doing with ReflectionUtils for my own projects. I use ver() a lot, so I statically import those.
Is `Collectors` used a lot in these classes? The answer is no.

PR #205, opened 2023-02-23
Fix setting skull owner using UUID pre-1.13

poyrazinan:
Fixed old version skull issue. Changed id.toString() to Bukkit.getOfflinePlayer(id).getName() because old version doesn't support uuid to owner skin. It only supports player name.


CryptoMorin:
Thanks

PR #211, opened 2023-05-23
Split runnables from task in XParticle

HSGamer:
This attempts to split all logic from particle tasks to be Runnable & BooleanSupplier, so that it can be used in other ways of scheduling tasks, like those new Folia's schedulers ([BukkitScheduler's deprecation note](https://jd.papermc.io/folia/1.19/org/bukkit/scheduler/BukkitScheduler.html)).


CryptoMorin:
Very nice, thank you for your contribution.

PR #223, opened 2023-08-27
Fixed return value of CreatureSpawner#getSpawnedType is null

### LINKED ISSUE #221: [XitemStack] - Serialize CreatureSpawner error
I receive this error, because the return of CreatureSpawner.getSpawnedType() is null. 
Need a new check before getting for example the name().

image

xNoci:
This PR resolves #221


CryptoMorin:
thanks

Issue #226, opened 2023-09-25
NoSuchMethod issue on init


aematsubara:
**Description**
I got this issue in the latest minecraft version, which happens when I use ReflectionUtils.MINOR_NUMBER to check the server version (tho I use this utility for other stuff too).

**Version**
Spigot and paper 1.20.2 (tested with both softwares), using XSeries 9.5.0

```
[16:18:17] [Server thread/WARN]: java.lang.NoSuchMethodException: no such method: net.minecraft.server.network.PlayerConnection.a(Packet)void/invokeVirtual
[16:18:17] [Server thread/WARN]: 	at java.base/java.lang.invoke.MemberName.makeAccessException(MemberName.java:976)
[16:18:17] [Server thread/WARN]: 	at java.base/java.lang.invoke.MemberName$Factory.resolveOrFail(MemberName.java:1117)
[16:18:17] [Server thread/WARN]: 	at java.base/java.lang.invoke.MethodHandles$Lookup.resolveOrFail(MethodHandles.java:3649)
[16:18:17] [Server thread/WARN]: 	at java.base/java.lang.invoke.MethodHandles$Lookup.findVirtual(MethodHandles.java:2680)
[16:18:17] [Server thread/WARN]: 	at me.matsubara.roulette.util.xseries.ReflectionUtils.<clinit>(ReflectionUtils.java:240)
[16:18:17] [Server thread/WARN]: 	at me.matsubara.roulette.RoulettePlugin.onEnable(RoulettePlugin.java:76)
[16:18:17] [Server thread/WARN]: 	at org.bukkit.plugin.java.JavaPlugin.setEnabled(JavaPlugin.java:266)
[16:18:17] [Server thread/WARN]: 	at org.bukkit.plugin.java.JavaPluginLoader.enablePlugin(JavaPluginLoader.java:342)
[16:18:17] [Server thread/WARN]: 	at org.bukkit.plugin.SimplePluginManager.enablePlugin(SimplePluginManager.java:480)
[16:18:17] [Server thread/WARN]: 	at org.bukkit.craftbukkit.v1_20_R2.CraftServer.enablePlugin(CraftServer.java:548)
[16:18:17] [Server thread/WARN]: 	at org.bukkit.craftbukkit.v1_20_R2.CraftServer.enablePlugins(CraftServer.java:462)
[16:18:17] [Server thread/WARN]: 	at net.minecraft.server.MinecraftServer.loadWorld0(MinecraftServer.java:577)
[...]

PR #228, opened 2023-09-30
Fix issues for 1.20.2 (#226)

datatags:
This PR fixes the issue mentioned in #226 as well as an NPE that occurs when trying to use SkullUtils.

I recognize the contributing guidelines say that changes shouldn't be made around the time a new minecraft version is released, but it's been over a week since 1.20.2 dropped, and I heard CryptoMorin is busy, so here's this. I've tried to keep the changes small and sensible, but if more changes are needed or if you'd like to just rewrite the changes from scratch, let me know.

The main change here is that the packet send method has been abstracted from `ServerGamePacketListenerImpl` to `ServerCommonPacketListenerImpl`. This change unfortunately occurred in its entirety across a "bug fix" version, so I've modified the `VersionHandler` class to accept patch versions as well, using overloading.

The other change is to simply replace `null` with `""` when creating a `GameProfile`, since the `name` parameter is now null-checked. It doesn't seem to affect functionality.

I've tested these changes on 1.20.1 and 1.20.2, and briefly on 1.8.8, and all seemed fine.

### DIFF
--- a/src/main/java/com/cryptomorin/xseries/ReflectionUtils.java
+++ b/src/main/java/com/cryptomorin/xseries/ReflectionUtils.java
@@ -193,7 +193,7 @@ public final class ReflectionUtils {
                 /* 17 */ 1,//            \_!_/
                 /* 18 */ 2,
                 /* 19 */ 4,
-                /* 20 */ 0,
+                /* 20 */ 2,
         };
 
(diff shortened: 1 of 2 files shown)


CryptoMorin:
Very nice. I'm just going to add your explanation to the code. Thanks.

PR #231, opened 2023-10-05
Armor trims and unreachable code fix

ImSpooks:
- Added support for Armor Trims
- Changed attribute modifier to use doubles instead of integers
  - This has been a double ever since it was implemented in spigot 1.9
- Fixed unreachable code in XItemStack
  - the following code would never reach when the server is on version 1.17 or higher, same goes for the other if checks below: https://github.com/CryptoMorin/XSeries/blob/6cbc1fc226f2eb0463700626e9f5db58f96ba339/src/main/java/com/cryptomorin/xseries/XItemStack.java#L779
  - Same applies here: https://github.com/CryptoMorin/XSeries/blob/6cbc1fc226f2eb0463700626e9f5db58f96ba339/src/main/java/com/cryptomorin/xseries/XItemStack.java#L305

### DIFF
--- a/src/main/java/com/cryptomorin/xseries/XItemStack.java
+++ b/src/main/java/com/cryptomorin/xseries/XItemStack.java
@@ -300,60 +300,68 @@ public final class XItemStack {
                     view.set("unlimited-tracking", mapView.isUnlimitedTracking());
                 }
             }
-        } else if (supports(17)) {
-            if (meta instanceof AxolotlBucketMeta) {
-                AxolotlBucketMeta bucket = (AxolotlBucketMeta) meta;
-                if (bucket.hasVariant()) config.set("color", bucket.getVariant().toString());
-            }
-        } else if (supports(16)) {
-            if (meta instanceof CompassMeta) {
-                CompassMeta compass = (CompassMeta) meta;
-                ConfigurationSection subSection = config.createSection("lodestone");
-                subSection.set("tracked", compass.isLodestoneTracked());
-                if (compass.hasLodestone()) {
-                    Location location = compass.getLodestone();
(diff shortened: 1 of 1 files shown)


src/main/java/com/cryptomorin/xseries/XItemStack.java
```diff
@@ -768,6 +771,18 @@ public static ItemStack edit(@Nonnull ItemStack item,
                 }
             }
         } else {
+            if (supports(20)) {
+                if (meta instanceof ArmorMeta) {
+                    ArmorMeta armorMeta = (ArmorMeta) meta;
+                    if (config.isSet("trim")) {
+                        ConfigurationSection trim = config.getConfigurationSection("trim");
+                        TrimMaterial trimMaterial = Registry.TRIM_MATERIAL.get(NamespacedKey.minecraft(trim.getString("material")));
+                        TrimPattern trimPattern = Registry.TRIM_PATTERN.get(NamespacedKey.minecraft(trim.getString("pattern")));
```
CryptoMorin, line 788: Can you make the namespace configurable as well? `namespace:key` format should do. If no namespace is specified, just use the default minecraft. I don't know if this is a thing in resource packs, but it'll probably be useful for the future.

The `NamespacedKey.fromString(trim.getString(...), "minecraft")` seems to do the job.

CryptoMorin (changes requested):

src/main/java/com/cryptomorin/xseries/XItemStack.java
ImSpooks: Suggested changes have been made, I also added armor trims to the serialization process.

CryptoMorin:
Thank you so much

PR #232, opened 2023-10-22
Fix splash potion material on old versions

Golfing7:
Pre-flattening, matchXMaterial(ItemStack) matches splash potions to the SPLASH_POTION material type. Using parseItem() then results in a null item. This update will fix that.


CryptoMorin:
Instead of changing the byte it to short, just add an exception inside parseItem method for splash potions.

Golfing7:
Sorry for the delay. I've edited the commit and it should be better now!

CryptoMorin:
Nice, thanks!

PR #241, opened 2024-02-04
[XItemStack] XItemStack.edit() removes unbreakable state when not specified in the config

Rubenicos:
```java
// Unbreakable item
ItemStack item = new ItemStack(Material.DIAMOND_SWORD);
ItemMeta meta = item.getItemMeta();
meta.setUnbreakable(true);
item.setItemMeta(meta);

ConfigurationSection config = new YamlConfiguration(); // empty config
// The item is no longer unbreakable
ItemStack sameItem = XItemStack.edit(item, config, Function.identity(), null);
```


CryptoMorin:
You should use isSet() instead which handles values based on the YAML settings.

CryptoMorin:
Thank you!

Issue #254, opened 2024-05-15
[XSkull] - Inconsistent behavior in different server versions & online/offline servers based on usernames and UUIDs


Condordito:
**Description**

At the moment, applying textures does not work correctly in multiple versions. Each one has slightly different behaviors. This issue was previously discussed in another post, but no further details were given on the topic (#235)

#### 1.12 (Online Server)
When applying a meta to an item stack, it generates the textures only if its GameProfile has a defined name or its UUID is stored in the server cache. Compared to the rest, the textures are applied synchronously, before being placed in an inventory

![Online 1 12](https://github.com/CryptoMorin/XSeries/assets/48872538/fa66f2b1-16d0-4071-a9e9-ed2401c4930e)

<img width="1026" alt="Screenshot 2024-05-15 at 1 48 03 AM" src="https://github.com/CryptoMorin/XSeries/assets/48872538/11fb28d9-051d-46a7-a3dc-74ec3d24d6eb">

<img width="1004" alt="Screenshot 2024-05-15 at 1 47 42 AM" src="https://github.com/CryptoMorin/XSeries/assets/48872538/cbce3d98-2120-4478-8bb0-b52a2ba2bd5a">

#### 1.12.2 (Offline Mode)
Only textures applied via a cached UUID will be visible and stacked. None applied with a name are visible, but this is because `SkullUtils.applySkinFromName` skips these cases. They apply correctly otherwise

![Offline 1 12](https://github.com/CryptoMorin/XSeries/assets/48872538/e2ab7b75-ba84-475f-99df-828c28546171)

#### 1.19.4 (Online Mode)
Textures from offline users will be applied on the next tick, causing them not to stack when placed in an inventory. As in 1.12, users with a non-cached UUID are not visible

![Online 1 19](https://github.com/CryptoMorin/XSeries/assets/48872538/c799b76f-aa5a-4ddc-998c-8251b9269717)

<img width="915" alt="Screenshot 2024-05-15 at 2 22 42 AM" src="https://github.com/CryptoMorin/XSeries/assets/48872538/bcac5de2-f4e9-4122-ab3a-a1e64c141bda">

<img width="1162" alt="Screenshot 2024-05-15 at 2 22 33 AM" src="https://github.com/CryptoMorin/XSeries/assets/48872538/1e003a4d-e907-491e-be6e-c72b9502f075">

#### 1.19.4 (Offline Mode)
[...]

PR #256, opened 2024-05-22
Fix XSkull#applySkin with name or UUID inputs (#254)

Condordito:
<h4>Additions</h4>

- Add support for offline-mode servers, as textures might not cache correctly.
- Implement asynchronous texture fetching option.

<h4>Changes</h4>

- Remove MOJANG_SHA_FAKE_ID_ENUMERATOR to prevent item stacking issues after restarts (UUIDs are now generated using texture hashes).
- Store UUIDs in usercache.json and cache textures on the first request.
- Ensure cached textures remain consistent across plugin interactions.

<h4>Fixes</h4>

- UUID requests returning HTTP response code: 429.
- CraftMetaSkull complaining about missing serializedProfile field.

<br>
Screenshot 2024-05-22 at 11 07 10 AM


CryptoMorin:
I think we need to make an inner class that makes dynamically configurable requests like ParticleDisplay.
For example, we could make the logging optional and propogate the error instead, or add two sync and async methods to allow the user to decide whether they want to keep the request in the same thread or not (in case the item meta is already being built in an async thread from the user)

Or we could just use the XSkull class itself as instead of creating a new inner class.
```java
    public static Builder of(ItemStack item) { return new Builder(item); }
    public static final class Builder {
        private ItemStack item;
        private SkullMeta meta;
        private SkullValue value;
        private GameProfile defaultValue = DEFAULT_PROFILE; // Or maybe a default SkullValue instead.
        private boolean propogateError = false, showWarnings = true;

        public Builder propogateError() {
            this.propogateError = true;
            return this;
        }

        public Builder fromString(String string) {
            value = /* detect it */;
            return this;
        }

        public Builder fromUsername(String name) {
            // fromUUID, fromBase64 etc
            value = new SkullValue(ValueType.NAME, name);
            return this;
        }

        public SkullMeta apply() {
            return applySkin(meta, value); // should make a separate method for this
        }

        public CompletableFuture<SkullMeta> applyAsync() {
            return CompletableFuture.supplyAsync(this::apply, PROFILE_EXECUTOR);
        }
    }
```

Condordito:
@CryptoMorin 

I made it more configurable and moved it to a separate package. It now logs exceptions using the debug level. So far, I've tested it on 1.8.9, 1.12.2, 1.19.4, and 1.20.4. I still need to document it and check for any additional remaps between those versions

src/main/java/com/cryptomorin/xseries/skull/XSkull.java
```diff
@@ -439,6 +631,13 @@ private static GameProfile fetchProfile(GameProfile profile) {
...
+    /**
+     * Sanitizes the provided {@link GameProfile} by removing unnecessary timestamp data
+     * and caches the sanitized profile.
+     *
+     * @param profile The {@link GameProfile} to be sanitized.
+     * @return The sanitized {@link GameProfile}.
+     */
     @SuppressWarnings("deprecation")
     private static GameProfile sanitizeProfile(GameProfile profile) {
         return Optional.ofNullable(Iterables.getFirst(profile.getProperties().get("textures"), null))
```
CryptoMorin, line 643: I don't think that the functional approach really suits this situation here. I prefer the traditional imperative style here.

src/main/java/com/cryptomorin/xseries/skull/XSkull.java
```diff
@@ -480,19 +688,33 @@ private static String extractTextureHash(String input) {
...
 
     /**
      * Tries to decode the string as a Base64 value.
-     * @return the decoded Base64 string if it is a Base64 string.
+     *
+     * @param base64 The Base64 string to decode.
+     * @return An {@link Optional} containing the decoded Base64 string if it is a valid Base64 string, or empty if not.
      */
-    private static Optional<String> decodeBase64(@Nonnull String base64) {
+    private static Optional<String> decodeBase64(String base64) {
```
CryptoMorin, line 717: Same as the previous code, specially since this is an internal code, it shouldn't use optionals. Just return null.

src/main/java/com/cryptomorin/xseries/skull/XSkull.java
```diff
@@ -424,6 +610,12 @@ private static void cacheProfile(GameProfile profile) {
         }
```
CryptoMorin, line 610: I've noticed you're using a lot of LOGGER.debug, you should just change them all to LOGGER.error that's definitely more suitable.

src/main/java/com/cryptomorin/xseries/skull/InputType.java
```diff
@@ -5,24 +5,61 @@
...
 
+    /**
+     * Returns the corresponding {@code InputType} for the given identifier, if it matches any pattern.
+     *
+     * @param identifier The string to be checked against the patterns.
+     * @return The matching {@code InputType}, or {@code null} if no match is found.
+     */
     @Nullable
     public static InputType get(@Nonnull String identifier) {
         return Arrays.stream(InputType.values())
```
CryptoMorin, line 61: You should definitely cache `InputType.values()`.

src/main/java/com/cryptomorin/xseries/skull/Action.java
```diff
@@ -22,19 +33,48 @@ public Thread newThread(final @Nonnull Runnable run) {
         }
```
CryptoMorin, line 33: Same here as XSkull, don't use LOGGER.debug, use LOGGER.error

src/main/java/com/cryptomorin/xseries/skull/XSkull.java
```diff
@@ -222,53 +222,115 @@ public final class XSkull {
...
+     * @param input The input string to retrieve the profile for.
+     * @return The {@link GameProfile} corresponding to the input type and input string.
+     *         Returns a profile with an invalid skull value if the type is {@code null}.
+     * @throws NullPointerException if {@code input} is {@code null}.
+     * @throws AssertionError if the input type is unknown.
+     */
     @Nonnull
     public static GameProfile getProfile(@Nullable InputType type, @Nonnull String input) {
+        Objects.requireNonNull(input, "Input cannot be null");
         if (type == null) return profileFromBase64(INVALID_SKULL_VALUE);
```
CryptoMorin, line 334: You should make all the profileFrom...() methods as a abstract method of InputType enum instead. That way this entire method would be useless as well.

src/main/java/com/cryptomorin/xseries/skull/Action.java
```diff
@@ -9,9 +9,20 @@
 import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.atomic.AtomicInteger;
 
+/**
+ * The {@code Action} class represents an action that can be performed on a {@link GameProfile}.
+ * It uses an {@link Instruction} to define how the action is applied.
+ *
+ * @param <T> The type of the result produced by the action.
+ */
 public class Action<T> {
```
CryptoMorin, line 18: Is there a point in having Instruction and Action as separate classes? We're only going to be using it for XSkull.
Also just for the sake of IDE tab-completion clutter, you should prefix all these new classes with `Skull`

src/main/java/com/cryptomorin/xseries/skull/XSkull.java
```diff
@@ -222,53 +222,115 @@ public final class XSkull {
...
+     *
+     * @param stack The {@link ItemStack} to set the profile for.
+     * @return An {@link Instruction} that sets the profile for the given {@link ItemStack}.
+     */
     public static Instruction<ItemStack> of(ItemStack stack) {
-        return new Instruction<>(
-            () -> XSkull.getProfile(stack),
-            (profile) -> XSkull.setProfile(stack, profile)
-        );
+        return new Instruction<>((profile) -> XSkull.setProfile(stack, profile));
```
CryptoMorin, line 232: We also need a method to create one from scratch:

```java
public static Instruction<ItemStack> create()
```

src/main/java/com/cryptomorin/xseries/skull/Instruction.java
```diff
@@ -3,33 +3,60 @@
...
     }
+
+    /**
+     * Sets the input type and value to be used in the profile setting operation and returns a new {@link Action} instance.
+     *
+     * @param type The type of the input value.
+     * @param input The input value to be used in the profile setting operation.
+     * @return A new {@link Action} instance configured with this {@code Instruction}.
+     */
     public Action<T> profile(InputType type, String input) {
```
CryptoMorin, line 57: We need individual methods for. `fromUUID(UUID)`, `fromPlayer(OfflinePlayer)`, `fromGameProfile(GameProfile)`
for other string-based types we can keep this method.
That of course means we need to refactor this class to support different types other than strings.

CryptoMorin:
Thank you very much! I'll have to test this a little and see if more changes are needed.

PR #259, opened 2024-06-05
[XParticle] BARRIER, LIGHT -> BLOCK_MARKER (v1.18)

CJCrafter:
In MC v1.18, the removed `BARRIER` and `LIGHT` particles in favor of the `BLOCK_MARKER` particle. More info here: https://minecraft.fandom.com/wiki/Particles

I feel that it is appropriate to have barrier and light as alternatives... though that is slightly opinionated since the new version of the particle requires block data to be sent.


CJCrafter:
https://hub.spigotmc.org/stash/projects/SPIGOT/repos/bukkit/diff/src/main/java/org/bukkit/Particle.java?until=27bcdd12f74d23dbc87265d78282c40147d0fd11

CryptoMorin:
Thanks!

PR #260, opened 2024-06-05
[XParticle] add BLOCK_CRACK alt for BLOCK particle

CJCrafter:
The `XParticle.BLOCK` enum is missing `BLOCK_CRACK` as an alt. `BLOCK_CRACK` was removed in 1.20.5 in favor of `BLOCK`

### DIFF
--- a/src/main/java/com/cryptomorin/xseries/particles/XParticle.java
+++ b/src/main/java/com/cryptomorin/xseries/particles/XParticle.java
@@ -71,7 +71,10 @@ public enum XParticle {
     ITEM_SLIME,
     HEART,
     ITEM,
-    BLOCK,
+    /**
+     * BLOCK_CRACK -> BLOCK (v1.20.5)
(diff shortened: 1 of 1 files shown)


CJCrafter:
Also here: https://hub.spigotmc.org/stash/projects/SPIGOT/repos/bukkit/diff/src/main/java/org/bukkit/Particle.java?until=8a34e009148cc297bcc9eb5c250fc4f5b071c4a7

I'll be making another PR in an hour or so with all of these enums^. I just want this one for my plugin, but I'll make a PR for the rest of them, too.

CryptoMorin:
You should've just added all these changes in a single PR, but it's fine.

CJCrafter:
> You should've just added all these changes in a single PR, but it's fine.

Yeah, I didn't realize at the time that I would make so many changes. Next time, I'll give it a few hours of thought

PR #261, opened 2024-06-06
[XParticle] v1.20.5 Renaming fix

CJCrafter:
Goes along with #260... Basically, it seems like there are a bunch of mistakes in XParticle, so I looked at the [spigot commit](https://hub.spigotmc.org/stash/projects/SPIGOT/repos/bukkit/diff/src/main/java/org/bukkit/Particle.java?until=8a34e009148cc297bcc9eb5c250fc4f5b071c4a7) that made all the changes and fixed the XParticle mistakes.

I also added some docs, and did some minor reorganization.


CryptoMorin:
Thank you so much. I really rushed thru the enums when adding XParticle.

PR #265, opened 2024-06-07
[ParticleDisplay] Be generally more user-friendly

datatags:
**Description**
The note particles are a bit strange: they have 25 possible colors, which are indicated in the `dX`/`offsetX` parameter using a scaled value between 0 and 1 (and count must be 0 for this field to act as the color, otherwise it acts as an offset as usual.) This means that in order to spawn a specific color of note particle, I'm currently required to do something like this:
```java
display.withColor(new Color((noteColor / 24f) * 255, 0, 0)).withCount(0);
```
To make this more user-friendly, I've added a function `withNoteColor` that accepts the index of a note color to display, and does this calculation internally. (See [this](https://minecraft.wiki/w/Note_Block#Notes) wiki page for the possible colors).

There may be a better way to do this as well, I'm open to any feedback!

### DIFF
--- a/src/main/java/com/cryptomorin/xseries/particles/ParticleDisplay.java
+++ b/src/main/java/com/cryptomorin/xseries/particles/ParticleDisplay.java
@@ -174,8 +174,8 @@ public class ParticleDisplay implements Cloneable {
     private Location location, lastLocation;
     @Nonnull
     private Vector offset = new Vector();
-    @Nonnull
-    private Vector particleDirection = new Vector();
+    @Nullable
+    private Vector particleDirection;
     /**
      * The direction is mostly used for APIs to call {@link #advanceInDirection(double)}
      * instead of handling the direction in a specific axis.
@@ -599,7 +599,7 @@ public class ParticleDisplay implements Cloneable {
             section.set("offset", offset.getX() + ", " + offset.getY() + ", " + offset.getZ());
         }
 
-        if (!isZero(display.particleDirection)) {
+        if (display.particleDirection != null) {
             Vector direction = display.particleDirection;
             section.set("direction", direction.getX() + ", " + direction.getY() + ", " + direction.getZ());
         }
@@ -1495,10 +1495,10 @@ public class ParticleDisplay implements Cloneable {
      * @since 11.0.0
      */
     @Nonnull
-    public ParticleDisplay particleDirection(Vector particleDirection) {
+    public ParticleDisplay particleDirection(@Nullable Vector particleDirection) {
         this.particleDirection = particleDirection;
         // Particle directions require a nonzero speed or the direction won't do anything.
-        if (extra == 0) extra = 1;
+        if (particleDirection != null && extra == 0) extra = 1;
         return this;
     }
 
@@ -1527,7 +1527,7 @@ public class ParticleDisplay implements Cloneable {
      * @since 2.1.0
      */
     public boolean isDirectional() {
-        return !isZero(particleDirection);
+        return particleDirection != null;
     }
 
     /**
@@ -1588,28 +1588,24 @@ public class ParticleDisplay implements Cloneable {
 
         Object data = null;
         if (this.data != null) {
-            this.data = this.data.transform(this.particle);
-            if (this.data.isOffsetBased(this.particle)) {
-                double[] offsetValues = this.data.offsetValues();
-                spawnWithDataInOffset(particle, loc, offsetValues, null);
+            this.data = this.data.transform(this);
+            Vector offsetData = this.data.offsetValues(this);
+            if (offsetData != null) {
+                spawnWithDataInOffset(particle, loc, offsetData, null);
                 return loc;
             }
-            data = this.data.data(this.particle, this.extra);
+            data = this.data.data(this);
             // Checks without data or block crack, block dust, falling dust, item crack or if data isn't right type
             if (!particle.getDataType().isInstance(data)) data = null;
         }
 
-        if (!isZero(this.particleDirection)) {
-            double[] offsetData = {particleDirection.getX(), particleDirection.getY(), particleDirection.getZ()};
-            spawnWithDataInOffset(particle, loc, offsetData, data);
+        if (particleDirection != null) {
+            spawnWithDataInOffset(particle, loc, particleDirection, data);
             return loc;
         }
 
         // Nothing weird, just spawn the particles normally.
-        double offsetx = offset.getX();
-        double offsety = offset.getY();
-        double offsetz = offset.getZ();
-        spawnRaw(particle, loc, count, offsetx, offsety, offsetz, data);
+        spawnRaw(particle, loc, count, offset, data);
         return loc;
     }
 
@@ -1620,11 +1616,11 @@ public class ParticleDisplay implements Cloneable {
      *
      * @param offsetData the data that needs to go in the offset fields.
      */
-    private void spawnWithDataInOffset(Particle particle, Location loc, double[] offsetData, Object data) {
+    private void spawnWithDataInOffset(Particle particle, Location loc, Vector offsetData, Object data) {
         // If there is no offset and we only want a single particle, we don't actually need to do anything special.
         // Otherwise, we'll at least need to use a loop.
         if (isZero(offset) && count < 2) {
-            spawnRaw(particle, loc, 0, offsetData[0], offsetData[1], offsetData[2], data);
+            spawnRaw(particle, loc, 0, offsetData, data);
             return;
         }
         // Particles with a specific direction must be flagged with count = 0,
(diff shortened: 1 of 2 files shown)


CryptoMorin:
Wouldn't it be better to make our own NoteColor enum inside ParticleDisplay since bukkit doesn't seem to have one? But you'd have to learn a few more color names because they're really close together. Also adding awt's color instance to each of them can help the user from their IDE.

image

datatags:
25 colors does seem like a lot to be individually named, maybe I could add a function that accepts a [Note](https://hub.spigotmc.org/javadocs/spigot/org/bukkit/Note.html) and uses the color corresponding to that? Or maybe an enum that refers to them by their note names directly, like `G_1` and `B_FLAT_2` where the numbers are the octave?

My primary use case for this API is just to make sequential note colors, I'm not sure what other use cases would be important to cover here.

CryptoMorin:
I guess most people that are going to be using this will base their choice on the note instead of the color itself, so I think your idea of using `Note` class is better.

But I also think it'd be nice to get withColor() method to handle the closest approximate color too. I was thinking about handling that in the spawn method instead, because someone might change the particle after setting the color, but that's a little bit extra.

datatags:
Added `withNoteColor(Note)` and the ability to convert between note colors and regular colors. A flag is used to indicate whether the floats stored in data[] represent a note index or a real color so that it knows whether to do the conversion.

As I'm writing this, it occurred to me that it might be cleaner to store an instance of `Note` itself as the data field when we're dealing with a note color, though it might require a bit of the logic in `spawn(Location)` to be restructured.

Thoughts?

As I was testing this, I realized that there doesn't appear to be a way to color the `ENTITY_EFFECT` particle with ParticleDisplay, because it's colored using an instance of `org.bukkit.Color` in the data field, rather than using the offset fields. That's a separate issue, but maybe could influence how this PR should be implemented as well.

datatags:
Decided to make some bigger changes. The main goal of this PR is now to just make particles work more like one would expect if one wasn't already familiar with how convoluted the particle system is. Also, I've cleaned up some things so they should be easier to maintain if more particle data changes in the future. So, now we have:
- No more opaque `Object data` field. Instead, it's now the interface `ParticleData`, which has subclasses for different types of data (regular color, note color, block data, etc.)
- Added a new field, `particleDirection`, which is a Vector that specifies the direction of particles that support it. As a result, `isDirectional()` now checks whether `particleDirection` is zero instead of checking `count`.
- Fields `count` and `offset` are now handled manually when the Bukkit API can't do it for us. For example, previously you couldn't do multiple note particles of a specific color, or a note particle with a specific color and offset, in both cases because you had to set count to 0 to indicate that you want the offset fields to represent the color. Instead, ParticleDisplay will now use the API for what it can, but if the offset fields are being used for data, it will manually calculate an offset for each particle in a similar way to the Bukkit API, and use a loop to spawn the appropriate number of them.
- The `extra` field is now also used as the size field, because the only particle that needs a size field is `DUST`/`DUST_TRANSITION`, and the regular `extra` parameter has no effect on those anyway.
- The `offset` vector was marked `@Nullable`, but there was no way for it to become null, so I've changed it to `@Nonnull`. `isZero(Vector)` is now used to check whether it "exists" or not.

I've done my best to maintain compatibility with previous versions of ParticleDisplay. I think the only casualty is the `getData()` field, since the underlying data field has been changed.

As always, I'm open to any feedback!

src/main/java/com/cryptomorin/xseries/particles/ParticleDisplay.java
```diff
@@ -1794,4 +1833,221 @@ public Vector mul(Vector point) {
...
+         * and false if it uses the data field.
+         * Sometimes depends on the particle and version.
+         */
+        public default boolean isOffsetBased(XParticle particle) {
+            return false;
+        }
+
+        public default double[] offsetValues() {
+            return new double[]{0, 0, 0};
+        }
```
CryptoMorin, line 1849: I think using a `Vector` here is more appropriate. Firstly because it's a fixed 3-valued object and because you wouldn't need to pass it around from spawn, spawnWithDataInOffset, and spawnRaw methods. An array would help the performance just a tiny bit, it's not worth it.

src/main/java/com/cryptomorin/xseries/particles/ParticleDisplay.java
```diff
@@ -1533,91 +1570,93 @@ public Location spawn(double x, double y, double z) {
...
+            double dz = offsetz == 0 ? 0 : r.nextGaussian() * 4 * offsetz;
+            Location offsetLoc = cloneLocation(loc).add(dx, dy, dz);
+            spawnRaw(particle, offsetLoc, 0, offsetData[0], offsetData[1], offsetData[2], data);
         }
     }
 
+    /**
+     * Calls the appropriate spawnParticle method with the parameters given.
+     */
+    private void spawnRaw(Particle particle, Location loc, int count, double dx, double dy, double dz, Object data) {
```
CryptoMorin, line 1651: Just use a `Vector` here for the offset then define dx,dy,dz variables for each inside the method since they're used 3 times.

src/main/java/com/cryptomorin/xseries/particles/ParticleDisplay.java
```diff
@@ -1794,4 +1833,221 @@ public Vector mul(Vector point) {
...
+
+    public interface ParticleData {
+        /**
+         * Returns true if the data type uses the offset fields,
+         * and false if it uses the data field.
+         * Sometimes depends on the particle and version.
+         */
+        public default boolean isOffsetBased(XParticle particle) {
+            return false;
+        }
```
CryptoMorin, line 1845: I believe you should just return null for `offsetValues()` if it's not supported. This is kinda extra.

src/main/java/com/cryptomorin/xseries/particles/ParticleDisplay.java
```diff
@@ -1794,4 +1833,221 @@ public Vector mul(Vector point) {
...
+        public ParticleData transform(XParticle particle) {
+            if (particle == XParticle.NOTE) {
+                return asNoteColor();
+            }
+            return this;
+        }
+
+        public NoteParticleColor asNoteColor() {
+            return new NoteParticleColor(findNearestNoteColor(color));
+        }
```
CryptoMorin, line 1922: This small single line is used once, we can just inline it inside transform method.

src/main/java/com/cryptomorin/xseries/particles/ParticleDisplay.java
```diff
@@ -1794,4 +1833,221 @@ public Vector mul(Vector point) {
...
+         */
+        public default boolean isOffsetBased(XParticle particle) {
+            return false;
+        }
+
+        public default double[] offsetValues() {
+            return new double[]{0, 0, 0};
+        }
+
+        public Object data(XParticle particle, double extra);
```
CryptoMorin, line 1851: I think it's better to just accept a ParticleDisplay here. Newer particles might require other data in the future.

src/main/java/com/cryptomorin/xseries/particles/ParticleDisplay.java
```diff
@@ -1794,4 +1833,221 @@ public Vector mul(Vector point) {
...
+
+        /**
+         * If this data doesn't support the given particle type but can be
+         * converted to a type that does, this method should return the appropriate
+         * ParticleData for that particle type.
+         * Used for converting RGB particle data to note block particle data.
+         */
+        public default ParticleData transform(XParticle particle) {
+            return this;
+        }
```
CryptoMorin, line 1863: Same as data() method, I think we should use a ParticleDisplay param.

src/main/java/com/cryptomorin/xseries/particles/ParticleDisplay.java
```diff
@@ -1794,4 +1833,221 @@ public Vector mul(Vector point) {
...
+        @Override
+        public ParticleData transform(XParticle particle) {
+            if (particle == XParticle.NOTE) {
+                return this;
+            }
+            return asRGBColor();
+        }
+
+        public RGBParticleColor asRGBColor() {
+            return new RGBParticleColor(NOTE_COLORS[note]);
```
CryptoMorin, line 1998: Please inline this.

src/main/java/com/cryptomorin/xseries/particles/ParticleDisplay.java
```diff
@@ -144,14 +162,20 @@ public class ParticleDisplay implements Cloneable {
...
     public boolean force;
     @Nonnull
     private XParticle particle = DEFAULT_PARTICLE;
     @Nullable
     private Location location, lastLocation;
-    @Nullable
+    @Nonnull
     private Vector offset = new Vector();
+    @Nonnull
+    private Vector particleDirection = new Vector();
```
CryptoMorin, line 178: Can't we just make this null instead of checking isZero every time?

CryptoMorin (changes requested):
I really like this, I actually started doing this 3 months ago but scratched the idea because of the massive changes it needed. Thanks!

src/main/java/com/cryptomorin/xseries/particles/ParticleDisplay.java
datatags: Good point!

datatags:
Done, also fixed a few more issues I noticed

PR #269, opened 2024-06-14
[XSound] Re-add play methods with volume and pitch for convenience

stupxd:
To improve compatibility with projects that use this method a ton.


CryptoMorin:
Sure I guess we can keep that.

PR #270, opened 2024-06-16
Added matchXEntityType methods

PeachesMLG:
Added matchXEntityType methods to match with other classes


CryptoMorin:
Those methods are literally already there above the ones you added, they're just named `of()` you should update those instead.

PR #271, opened 2024-06-17
public XEntity#editBossBar

HSGamer:
I think this is useful for more cases than just as a private method for XEntity#edit


CryptoMorin:
sure

PR #274, opened 2024-06-20
[XSkull] Legacy support

### LINKED ISSUE #273: [XSkull] - Legacy support
**Description**
`ProfilesCore` will try to get `GameProfileCache#getNextOperation`, but this method/field is not present in older server versions. I don't think `MojangAPI#profileFromUsername0` should re-implement the whole method, as the server will already save and check for expired entries every time a user logs in (`PlayerList#placeNewPlayer`)

Screenshot 2024-06-20 at 7 58 10 AM

**Version**
Spigot 1.8.9, 1.12.2

Condordito:
This PR fixes #273 and some other minor bugs 

### Changes
- Fix `MojangAPI.MOJANG_PROFILE_CACHE` condition
- Fix `ProfilesCore.UserCacheEntry_getProfile` not being accessible
- Fix remaps for `UserCache_profilesByUUID` and `UserCache_profilesByName`
- Remove unused reflected methods/fields
- Remove `Profileable#of(Player)`
Screenshot 2024-06-20 at 9 24 37 AM
Screenshot 2024-06-20 at 9 26 12 AM
Screenshot 2024-06-20 at 9 28 14 AM

### DIFF
--- a/src/main/java/com/cryptomorin/xseries/profiles/ProfilesCore.java
+++ b/src/main/java/com/cryptomorin/xseries/profiles/ProfilesCore.java
@@ -32,15 +32,12 @@ public final class ProfilesCore {
 
     public static final Map<String, Object> UserCache_profilesByName;
     public static final Map<UUID, Object> UserCache_profilesByUUID;
-    public static final Deque<GameProfile> UserCache_gameProfiles;
 
     public static final MethodHandle
             FILL_PROFILE_PROPERTIES, GET_PROFILE_BY_NAME, GET_PROFILE_BY_UUID, CACHE_PROFILE,
             CRAFT_META_SKULL_PROFILE_GETTER, CRAFT_META_SKULL_PROFILE_SETTER,
             CRAFT_SKULL_PROFILE_SETTER, CRAFT_SKULL_PROFILE_GETTER,
-            PROPERTY_GET_VALUE,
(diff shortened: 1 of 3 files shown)


CryptoMorin:
Btw, I forgot this, but I'm not sure why you removed
`public static final Deque<GameProfile> UserCache_gameProfiles;`

That is present in older versions and is still considered a part of the cache.

I will add back `UserCacheEntry_setLastAccess` because we are still modifying a cache. You can't just remove this just because it doesn't exist in older versions.

Condordito:
@CryptoMorin 

> public static final Deque<GameProfile> UserCache_gameProfiles;
That is present in older versions and is still considered a part of the cache.

> I will add back UserCacheEntry_setLastAccess because we are still modifying a cache. You can't just remove this just because it doesn't exist in older versions.

`MojangAPI#cacheProfile` will already handle this by calling the server implementation

## 1.8.9
Screenshot 2024-06-22 at 1 39 51 AM

## 1.20.6 
Screenshot 2024-06-22 at 1 42 50 AM
Screenshot 2024-06-22 at 1 42 45 AM

CryptoMorin:
Oh I see, my bad. I added the Deque just to remove from it which is not needed anymore.
I'm not sure of any current side effects of setLastAccess() either, but since we're using the cache, let's just keep it consistent.

Issue #278, opened 2024-06-24
[XSkull] - Unable to set profile


thomasmny:
**Description**
Trying to create a skull and then adding it to a GUI in 1.20.4 using
```java
ItemStack skull = XSkull.createItem()
                .profile(Profileable.detect("d5c6dc2bbf51c36cfc7714585a6a5683ef2b14d47d8ff714654a893f5da622"))
                .lenient()
                .apply();
```
causes a `RuntimeExecption`. This does not occur in 1.20.6.

**Error log**
https://pastebin.com/yuc3dKg8

**Version**
Paper version git-Paper-497 (MC: 1.20.4) (Git: d8d54d9 on ver/1.20.4)
Note: Yes, I know 1.20.6 has been released, but some plugins such as Arceon don't have a 1.20.6 release yet

PR #279, opened 2024-06-24
[XSkull] Support 1.20.1-1.20.5 proxy mappings

rexlManu:
This pr fixes the mappings for the proxy. You can also confirm those mappings by https://mappings.cephx.dev/


rexlManu:
My pure coindicant this will also fix #278 since I had similar issues with 1.20.1 and I tested my plugin against 1.20.4+ and now it works again

CryptoMorin:
The code's format is a bit messed up, but it's fine, I'll fix it. Thank you!

PR #283, opened 2024-07-09
[XItemStack] Support hex colors

datatags:
In the XItemStack docs, `Color` is defined as:
```
alias Color: for Text "red, green, blue" or "hex color"
```
However, XItemStack actually doesn't currently parse hex colors, only the comma-separated format is supported. This PR changes that, and now all of the following formats are supported:
- `r, g, b` (just like before)
- `#RRGGBB`
- `0xRRGGBB`
- decimal representation of `r << 16 | g << 8 | b`, as used in NBT

### Notes

The hex representation is automatically converted to decimal by SnakeYAML (and then converted to a string), so that format and the decimal representation end up being handled the same in the code. The bare format `RRGGBB` is also supported if it can't be parsed as a base-10 number, but due to that restriction I don't recommend that.

### DIFF
--- a/src/main/java/com/cryptomorin/xseries/XItemStack.java
+++ b/src/main/java/com/cryptomorin/xseries/XItemStack.java
@@ -1137,6 +1137,11 @@ public final class XItemStack {
     /**
      * Parses RGB color codes from a string.
      * This only works for 1.13 and above.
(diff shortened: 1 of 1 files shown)


CryptoMorin:
Thanks!

PR #286, opened 2024-07-14
Add Jitpack settings

HSGamer:
This is for those who want to use dev-builds for quick testing and delivering.

It's still recommended to use the release build.

### DIFF
--- a/jitpack.yml
+++ b/jitpack.yml
@@ -1,3 +1,4 @@
 before_install:
    - sdk install maven
-   - sdk use maven
+   - sdk update
+   - sdk default maven
--- a/jitpack.yml
+++ b/jitpack.yml
@@ -1,4 +1,5 @@
+jdk:
+  - openjdk17
 before_install:
    - sdk install maven
-   - sdk update
    - sdk default maven
--- a/jitpack.yml
+++ b/jitpack.yml
@@ -1,0 +1,3 @@
+before_install:
+   - sdk install maven
+   - sdk use maven
--- a/jitpack.yml
+++ b/jitpack.yml
@@ -1,5 +1,5 @@
 jdk:
   - openjdk17
 before_install:
-   - sdk install maven
-   - sdk default maven
+  - sdk install maven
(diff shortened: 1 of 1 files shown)


CryptoMorin:
I'm sorry but how does this work exactly? Are you asking me to start publishing builds to JitPack?

HSGamer:
> I'm sorry but how does this work exactly? Are you asking me to start manually publishing builds to JitPack or something?
> You just added a jitpack setting file which is adding unnecessary information that [JitPack can derive from other settings](https://jitpack.io/docs/BUILDING/#java-version).
> 
> But for some reasons [JitPack is still using JDK 8](https://jitpack.io/com/github/CryptoMorin/XSeries/11.2.0.1/build.log):
> > Java version: 1.8.0_292, vendor: Private Build, runtime: /usr/lib/jvm/java-8-openjdk-amd64/jre
> 
> But the main issue is the outdated Maven version:
> > [ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.13.0:compile (default-compile) on project XSeries: The plugin org.apache.maven.plugins:maven-compiler-plugin:3.13.0 requires Maven version 3.6.3 -> [Help 1]
> 
> Are you suggesting that by manually configuring JDK 17, it will automatically use a higher Maven version?
> Some JitPack releases, builds and commits are fine, but some of them are randomly failing. I'm not sure why that's happening. The successful ones don't have a log so I can't tell.

First off, there are some logs suggesting that the maven plugins XSeries is using are not working with the Maven version Jitpack is using. So there are lines to install latest Maven manually (`before_install`).
Secondly, if we don't define `jdk` in the setting, Jitpack uses Java 8 by default and it'll not work. So we need to set that manually.
No, jdk is not enough to update Maven. We need other lines to install it manually, hence the `sdk` lines.

And I don't ask you to start publishing on Jitpack. I just added the required settings for those who want to test the commits on Jitpack, to finally build it successfully.

PR #299, opened 2024-09-18
[XSkull] Support v1.21.1b78+ ResolvableProfile

### LINKED ISSUE #298: [XSkull] - New "public ResolvableProfile profile" field of CraftMetaSkull causes reflection issues on v1.21.1b78+
**Description**
XSkull has once again encountered a reflection error, occurring on the most recent server builds for 1.21.1. Stack trace: https://pastebin.com/xUdChkPb

**Version**
XSeries 11.2.0
Happens on Spigot (`CraftBukkit version 4316-Spigot-a759b62-8a6f8b6 (MC: 1.21.1)`) and Paper (`Paper version 1.21.1-81-master@13a2395 (2024-09-15T20:39:09Z)`)

rexlManu:
Already sometime the ResolvableProfile exists (I think since 1.20.5) but now the GameProfile was replaced with the ResolvableProfile both in the CraftSkull & CraftMetaSkull classes. ResolvableProfile has a constructor accepts only a GameProfile to create a profile from a GameProfile and it also saves a gameProfile that is also created when it's not created with a game profile.

This pr adds two new method handles. One for creating a ResolvableProfile from a GameProfile and one to get the GameProfile back from a ResolvableProfile. I tested myself the changes for my plugins and with my changes it, Skulls (both Items & Blocks) working again on paper build 83.

This would also fix #298

_If I'm honest, I have zero clue how your abstraction over reflection even works and it looks really "magically" and I'm sure I didn't properly followed the current code style, so if you want to implement yourself, just use this pr to see how it works._

### DIFF
--- a/src/main/java/com/cryptomorin/xseries/profiles/ProfilesCore.java
+++ b/src/main/java/com/cryptomorin/xseries/profiles/ProfilesCore.java
@@ -65,7 +65,14 @@ public final class ProfilesCore {
             MinecraftClassHandle ResolvableProfile =
                     ns.ofMinecraft("package nms.world.item.component; public class ResolvableProfile {}");
 
-            boolean useResolvableProfile = CraftMetaSkull.field("private ResolvableProfile profile;").getter().exists();
+            boolean useResolvableProfile;
+            try {
+                CraftMetaSkull.field("private ResolvableProfile profile;").getter().exists();
+                useResolvableProfile = true;
+            }catch (Exception e) {
+                // catch Unknown type 'ResolvableProfile' -> 'ResolvableProfile' if type is GameProfile
+                useResolvableProfile = false;
+            }
 
             if(useResolvableProfile) {
                 newResolvableProfile = ResolvableProfile.constructor("public ResolvableProfile(GameProfile gameProfile);").reflect();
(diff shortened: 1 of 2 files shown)


CryptoMorin:
Thank you so much. The way you used the reflection is actually correct. There are just small semantic issues, but that's fine, I can change those later.

Are you able to test this in v1.21.1 build 50, v1.21.0, and v1.20.6 just to be sure they work correctly?

rexlManu:
Alright I tested it with paper build 50, 1.21 (doesn't exists anymore since it was replaced with 1.21.1 but still tested it tho), 1.20.6 and 1.20.4. I found a issue on 1.20.4 and older and changed how I check the type. I assumed ReflectiveHandle#exists, covers the exception but the exception is still thrown. I replaced it with a inner try catch. I don't like that approach but I'm just to unfamiliar with the reflection abstraction, so that needs abit of touch.

rexlManu:
What I should mention is that I only tested it with paper. The only different for spigot is that the gameProfile() method is [called f() for now (1.20.6-1.21.1)](https://mappings.dev/1.21.1/net/minecraft/world/item/component/ResolvableProfile.html)

PR #301, opened 2024-09-28
[XReflection] - ReflectionParser usage of toUpperCase with Locale.ENGLISH

poyrazinan:
Added Locale.ENGLISH for parse correct format.

An error occurred when try to parse PUBLIC in Turkish like PUBLİC


CryptoMorin:
Thanks I forgot that.

PR #304, opened 2024-10-13
[XMaterial] Fix NPE when no server is running (during unit tests)

Almighty-Satan:
My project has some unit tests that run without starting a Bukkit server. This causes a NullPointerException in `XMaterial` when `Bukkit#getVersion` is called:
```
Caused by: java.lang.NullPointerException: Cannot invoke "org.bukkit.Server.getVersion()" because "org.bukkit.Bukkit.server" is null
	at org.bukkit.Bukkit.getVersion(Bukkit.java:96)
	at com.cryptomorin.xseries.XMaterial$Data.<clinit>(XMaterial.java:2317)
	... 9 more
```
The tests don't actually call any `XMaterial` methods, but the class is initialized by an enum that looks something like this:
```java
VALUE_1("Some string", XMaterial.FURNACE),
VALUE_2("Some other string", XMaterial.CLOCK)
```
This PR aims to fix this by checking if `Bukkit#getServer` is `null` before calling `Bukkit#getVersion`

### DIFF
--- a/src/main/java/com/cryptomorin/xseries/XMaterial.java
+++ b/src/main/java/com/cryptomorin/xseries/XMaterial.java
@@ -2314,11 +2314,15 @@ public enum XMaterial /* implements com.cryptomorin.xseries.abstractions.Materia
         private static final int VERSION;
 
         static { // This needs to be right below VERSION because of initialization order.
-            String version = Bukkit.getVersion();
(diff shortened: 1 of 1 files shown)


CryptoMorin:
Sure I guess we can use this, but adding a `System.err.println` for this might be appropriate just in case something weird actually happens.

Issue #319, opened 2024-12-28
[XBiome] World#getMinHeight check sets wrong boolean (maxHeight) to true


SpraxDev:
Hi, it looks like the #getMinHeight support-check sets the maxHeight-boolean flag to true instead of the minHeight one.

This looks like an accidental copy'n paste error that might result in unexpected behavior in some server versions maybe?

https://github.com/CryptoMorin/XSeries/blob/b51d9fdce75529a9d39f596cffe46c8ce1e72dc0/src/main/java/com/cryptomorin/xseries/XBiome.java#L174-L192

(line 187)

CryptoMorin:
Oh, yeah. I didn't notice that.

PR #320, opened 2025-01-01
[XBiome] Fix World#getMinHeight Check Setting Wrong Boolean

IanTapply22:
Fixes: #319

Fix a typo of setting maxHeight rather than the minHeight value in the World#getMinHeight check.


CryptoMorin:
Thanks

PR #321, opened 2025-01-04
expose the form classes

CJCrafter:
Having this data solves some minor problems for me. Being able to get the types is good when we want to wrap your generics with our own generics (if we do anything fancy). 

This PR just adds getters for the types already stored in `XRegistry`

### DIFF
--- a/src/main/java/com/cryptomorin/xseries/base/XRegistry.java
+++ b/src/main/java/com/cryptomorin/xseries/base/XRegistry.java
@@ -150,6 +150,20 @@ public final class XRegistry<XForm extends XBase<XForm, BukkitForm>, BukkitForm>
     }
 
     /**
+     * Gets the class of the bukkit form.
+     */
+    public Class<BukkitForm> getBukkitFormClass() {
+        return bukkitFormClass;
+    }
+
+    /**
+     * Gets the class of the xform.
+     */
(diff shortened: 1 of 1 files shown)


CryptoMorin:
I'm ok with merging this, but can you maybe give a practical example of how this can be used? I don't see how this helps with generics.

CJCrafter:
Yeah ofc! Here is an example I'm working in (sorry for the Kotlin)

In my serializers for config deserializer, I have a bunch of "`String -> <T>`" serializers... Here is one of them:
```kotlin
class XSeriesValueSerializer<XForm : XModule<XForm, BukkitForm>, BukkitForm>(
    private val registry: XRegistry<XForm, BukkitForm>,
    private val isAllowWildcard: Boolean,
) {
    override fun getSerializedType(): Class<BukkitForm> = registry.getBukkitFormClass()

    // ...
}
```

I need this to:
1. cast objects later (`clazz.cast` makes me happy over `(T)`)
2. show users a simple class name of the types

---

I can get around that by passing the class(es) in the constructor, or iterating over the registry and grabbing the types from the first iteration. These workarounds are just a bit gross.

CryptoMorin:
Oh I see. Yeah I can see that being useful in rare cases like these. (I use Kotlin extensively for my other projects too, so no worries 😄)

PR #323, opened 2025-01-10
[XBlock] XBlock.isSimilar() Inconsistent pre-1.13 and post-1.13 Material enum checks

stupxd:
This should fix issues with attempting to access enum values from newer versions in 1.8, also false positives when using direct check (e.g. `XBlock.isMelon(null)`), as it would return true in 1.13+ because `BlockMaterial.MELON_BLOCK.material` is `null`.


CryptoMorin:
Thanks! Just a small change, can you replace all the `XMaterial.supports(...)` with the `ISFLAT` static bool?

stupxd:
Oh yeah sure, will do today

PR #331, opened 2025-03-17

Devlrxxh:


CryptoMorin:
These are not "small changes" you literally changed the Maven settings. Why?

Devlrxxh:
Sorry I didn't mean to push that I will create a new pr

PR #332, opened 2025-03-18

Devlrxxh:


Devlrxxh:
I also tired this in game and it fixed my issue

PR #333, opened 2025-03-18

Devlrxxh:
I also tired this in game and it fixed my issue

### DIFF
--- a/core/src/main/java/com/cryptomorin/xseries/XMaterial.java
+++ b/core/src/main/java/com/cryptomorin/xseries/XMaterial.java
@@ -1835,14 +1835,12 @@ public enum XMaterial implements XBase<XMaterial, Material> {
         this.legacy = legacy;
 
         Material mat = Data.getExactMaterial(name());
-        if (mat == null) {
-            if ((!Data.ISFLAT && this.isDuplicated())) {
+            if ((!Data.ISFLAT && this.isDuplicated()) || mat == null) {
                 for (int i = legacy.length - 1; i >= 0; i--) { // Backwards checkup
                     mat = Data.getExactMaterial(legacy[i]);
                     if (mat != null) break;
                 }
             }
-        }
 
         this.material = mat;
     }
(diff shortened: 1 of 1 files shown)


CryptoMorin:
If you provide a minimal code to reproduce this issue like I asked it'd be really helpful...
And the server version you tested this on.

Devlrxxh:
```
player.getInventory().addItem(XMaterial.RED_DYE.parseItem());
```
Results:
 1.21 Paper: Returns INC_SAC instead of RED_DYE (As that's the legacy value)

1.8 Paper: Returns the correct item stack

CryptoMorin:
> 1.21 Paper: Returns INC_SAC instead of RED_DYE (As that's the legacy value)

My code:
```java
log("RED_DYE is " + XMaterial.RED_DYE.parseItem());
```

Output:
> RED_DYE is ItemStack{RED_DYE x 1}

Version:
> This server is running CraftBukkit version git-Spigot-73860e0-f0c8dd3 (MC: 1.21.4) (Implementing API version 1.21.4-R0.1-SNAPSHOT)

Same with v1.21.1

Devlrxxh:
Yeah that's how it's supposed to be also that's red dye you got on 1.21.4

Devlrxxh:
Before the change I always got a INC instead

Devlrxxh:
When testing it, I get INC_SAC try this `XMaterial.matchXMaterial("RED_DYE").get().parseItem()`

CryptoMorin:
> When testing it, I get INC_SAC try this `XMaterial.matchXMaterial("RED_DYE").get().parseItem()`

Still works for me `ItemStack{RED_DYE x 1}` even tested in the new v1.21.5

CryptoMorin:
If you mean `api-version` like mentioned in the [main page](https://github.com/CryptoMorin/XSeries), yes that can sometimes affect the values. You need to add that.

Devlrxxh:
ah okay, thanks a lot

PR #335, opened 2025-03-24
[XMaterial] Change WHITE_DYE and BLACK_DYE for legacy versions

fleivinho:
Resolving these materials that return null in versions like 1.8.X. My item system turns null materials into stone, and these items becoming stone is problematic.


CryptoMorin:
`WHITE_DYE` is not a `15, "INK_SACK"` that's for BONE_MEAL.
`BLACK_DYE` is not a `0, "INK_SACK"` that's for INK_SAC itself. INK_SACK and BLACK_DYE are two different materials.

PR #336, opened 2025-03-29
[ParticleDisplay] Ensure 'extra' isn't 0 when setting note color

datatags:
You would expect this code to produce a purple music note (according to the table [here](https://minecraft.wiki/w/Note_Block#Notes)):
```java
ParticleDisplay.of(XParticle.NOTE).withColor(12).spawn(loc);
```
But as it stands, you only get green, no matter what color you put in, due to `extra` defaulting to 0. For whatever reason, `extra` is multiplied by the first offset field (the place where the note color is typically stored) and so if either is 0, you only get green notes back. So the working code is this:
```java
ParticleDisplay.of(XParticle.NOTE).withColor(12).withExtra(1).spawn(loc);
```

That's definitely unexpected if you're not already used to the weirdness of the particle system, and perhaps even if you are. So, to resolve confusion, I've added a check to ensure `extra` is not 0 when setting the note color. This parallels what ParticleDisplay already does in `particleDirection` and `spawnRaw` for dust particles.

### DIFF
--- a/core/src/main/java/com/cryptomorin/xseries/particles/ParticleDisplay.java
+++ b/core/src/main/java/com/cryptomorin/xseries/particles/ParticleDisplay.java
@@ -1644,9 +1644,14 @@ public class ParticleDisplay {
         double dx = offset.getX();
         double dy = offset.getY();
         double dz = offset.getZ();
-        // The "extra" field has no effect on dust particles in some versions,
-        // but in others it causes the colors to not display when set to 0.
-        double extra = (this.particle == XParticle.DUST) ? 1 : this.extra;
+        double extra = this.extra;
+        if (this.particle == XParticle.DUST || this.particle == XParticle.NOTE) {
+            // The "extra" field has no effect on dust particles in some versions,
+            // but in others it causes the colors to not display when set to 0.
+            // For note particles, the rendered color is dx * extra, so we just
+            // always put the color in dx and set extra to 1.
+            extra = 1;
+        }
         if (players == null)
             if (ISFLAT)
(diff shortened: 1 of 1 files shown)


CryptoMorin:
That's interesting. Just a small detail:

The example shown uses `withColor()` but the PR changes `withNoteColor()`, and `withColor()` data [automatically transforms itself into **NoteParticleColor**](https://github.com/CryptoMorin/XSeries/blob/81bb82a0d56cfe7d03a3a2658dc25f1c42de0267/core/src/main/java/com/cryptomorin/xseries/particles/ParticleDisplay.java#L1910-L1916) when **XParticle.NOTE** is used, however during the transformation `this.extra` isn't changed. This practically works when `withColor()` is used without an explicit size parameter because the default size is 1. And of course the note color doesn't have different sizes, so it can still be error-prone. Maybe we can add a check for this and throw exceptions when `withColor(color, size)` is used directly on a note particle? `withParticle()` is still going to be problematic if its used after `withColor(color, size)` but that method is just error-prone in general for other data as well. We can fix that later.

datatags:
Oops, I meant to put `withNoteColor()` in the examples, but that's a good point as well. Maybe it'd be easier to just force `extra` to always be 1 when the particle is spawned, like we already do [for dust particles](https://github.com/CryptoMorin/XSeries/blob/4cc0439cd6dfa78bd22343f47862aef4957e42e3/core/src/main/java/com/cryptomorin/xseries/particles/ParticleDisplay.java#L1652)?

CryptoMorin:
that could work yeah

CryptoMorin:
Thanks!

PR #359, opened 2025-07-22
[XMaterial] Fix CARVED_PUMPKIN material on versions below 1.13

radek203:


CryptoMorin:
It's interesting that carved pumpkins were a thing before normal pumpkins. Thanks!

PR #393, opened 2025-12-29
[XGameRule] Add XGameRule

### LINKED ISSUE #386: [XGameRule] Add XGameRule
The gamerule name was recently changed in version 1.21.11

Image

Almighty-Satan:
This implementation is unfinished, but I wanted to get some feedback before putting more work into this.

I decided to use Strings instead of `org.bukkit.GameRule` because I need this to work on 1.8.8 and that class does not exist on that version (I think it was added around 1.14). Therefore `XGameRule` was made an enum and does not use `XRegistry`.

Example Usage:  
```java
World world = ...;
world.setGameRuleValue(XGameRule.ADVANCE_TIME.get(), "false");
```

Closes #386

### DIFF
--- a/core/src/main/java/com/cryptomorin/xseries/XGameRule.java
+++ b/core/src/main/java/com/cryptomorin/xseries/XGameRule.java
@@ -187,7 +187,7 @@ public final class XGameRule<T> implements XBase<XGameRule<T>, String> {
     private final String[] names;
     private final Class<?> type;
     private final String value;
-    private Object rule;
+    private final Object rule;
 
     private XGameRule(@NotNull Class<T> type, @NotNull String... names) {
         this.names = names;
@@ -196,12 +196,10 @@ public final class XGameRule<T> implements XBase<XGameRule<T>, String> {
         World world = Bukkit.getWorlds().get(0);
         this.value = Arrays.stream(names).filter(world::isGameRule).findAny().orElse(null);
 
-        if (SUPPORTS_GameRule && this.value != null) {
-            try {
-                this.rule = GameRule_getByName.invoke(this.value);
-            } catch (Throwable e) {
-                throw new RuntimeException(e);
-            }
+        try {
+            this.rule = SUPPORTS_GameRule && this.value != null ? GameRule_getByName.invoke(this.value) : null;
+        } catch (Throwable e) {
+            throw new RuntimeException(e);
         }
     }
 


CryptoMorin:
I'm not sure how this is supposed to be cross-compatible? How do I set a GameRule in v1.21.11 with this class? `world.setGameRuleValue` doesn't exist now. Also, your names are inconsistent, I don't know if you've seen the XAttribute class, but newer names must appear first in each name list. It's just a coding convention.

Also, you need to conform to the latest features to include the most information. The latest [GameRule](https://hub.spigotmc.org/javadocs/bukkit/org/bukkit/GameRule.html) API uses separate fields with generic information. You need to do the same for each enum. You'd probably have to do that for 1.21.11+ support anyway.

core/src/main/java/com/cryptomorin/xseries/XGameRule.java
```diff
@@ -139,15 +143,61 @@ public enum XGameRule implements XBase<XGameRule, String> {
     @XInfo(since = "?", removedSince = "1.21.11")
     DO_FIRE_TICK("doFireTick");
 
+    private static final boolean SUPPORTS_GAME_RULE_API;
+    private static final MethodHandle GET_BY_NAME;
+    private static final MethodHandle GET_GAME_RULE_VALUE;
```
CryptoMorin: -> World_getGameRuleValue

core/src/main/java/com/cryptomorin/xseries/XGameRule.java
```diff
@@ -139,15 +143,61 @@ public enum XGameRule implements XBase<XGameRule, String> {
...
+        try {
+            getByName = methodHandles.unreflect(GameRule.class.getDeclaredMethod("getByName", String.class));
+        } catch (NoSuchMethodException | IllegalAccessException e) {
+            throw new RuntimeException(e);
+        } catch (NoClassDefFoundError e) {
+            supportsGameRuleAPI = false;
+
+            try {
+                getGameRuleValue = methodHandles.unreflect(World.class.getDeclaredMethod("getGameRuleValue", String.class));
+                setGameRuleValue = methodHandles.unreflect(World.class.getDeclaredMethod("setGameRuleValue", String.class, String.class));
```
CryptoMorin: Can't we just use the lookup directly? Why have the reflection intermediate?

```java
lookup.findVirtual(World.class, "setGameRuleValue", MethodType.methodType(void.class, String.class, String.class));
```

It really doesn't really matter, with direct findVirtual lookups the startup performance and memory usage might improve which is negligible, it's mostly a matter of convention. I'd appreciate if you can fix the other 2 as well.

core/src/main/java/com/cryptomorin/xseries/XGameRule.java
```diff
@@ -157,4 +207,44 @@ public String[] getNames() {
...
+        if (SUPPORTS_GAME_RULE_API) {
+            if (this.rule == null)
+                throw new UnsupportedOperationException("Game rule not supported on this version!");
+
+            @SuppressWarnings("unchecked")
+            GameRule<T> rule = (GameRule<T>) this.rule;
+            return world.setGameRule(rule, value);
+        } else {
+            if (this.value == null)
+                throw new UnsupportedOperationException("Game rule not supported on this version!");
```
CryptoMorin: Duplicated check, move out of if-branch.

core/src/main/java/com/cryptomorin/xseries/XGameRule.java
```diff
@@ -157,4 +207,44 @@ public String[] getNames() {
...
+        if (SUPPORTS_GAME_RULE_API) {
+            if (this.rule == null)
+                throw new UnsupportedOperationException("Game rule not supported on this version!");
+
+            GameRule<?> rule = (GameRule<?>) this.rule;
+            Object value = world.getGameRuleValue(rule);
+            return value == null ? null : value.toString();
+        } else {
+            if (this.value == null)
+                throw new UnsupportedOperationException("Game rule not supported on this version!");
```
CryptoMorin: Duplicated check, move out of if-branch.

CryptoMorin:
> Is the generic information really needed?

I mean, they all refer to different types, no? From what I can see, you're just converting them all to string. I don't think that's a good idea. You can throw the enum into an AI and tell it to generate a field for you. You can even take a step further and tell it to fetch the correct generic information automatically and even add summary comments for each field about what the gamerule does.

The `UnsupportedOperationException` checks are repeated twice, I also marked them. Use `if (!isSupported())` for them too just to be consistent.

And lastly, I usually name my MethodHandles like `<class>_<method>`
I'll mark them in the review.

Sorry if all this seems a bit too nit-picky, I just want to keep it consistent.

CryptoMorin (changes requested):

Almighty-Satan:
I've changed XGameRule from an enum to a class so the constants can have a generic type that can be used as a return type for `XGameRule#getValue`.

CryptoMorin:
Thank you!

PR #412, opened 2026-06-19
[XTag] Add missing armor and tool tags

iiAhmedYT:


iiAhmedYT:
The force-push was to reorder the missing items to follow one order incase some plugins use it to display their items

CryptoMorin:
Thanks!

PR #419, opened 2026-08-17
[XSkull] usercache.json corruption on proxied (BungeeCord/Velocity) backends

timderspieler:
### The problem

On backends behind an online-mode proxy (Velocity/BungeeCord with player-info forwarding), XSkull writes wrong entries into the server's `usercache.json`: profiles it resolves are stored under the **computed offline UUID** (`UUID.nameUUIDFromBytes("OfflinePlayer:" + name)`), even though every player on such a server has their **real Mojang UUID** (the proxy authenticates and forwards it).

This was reported against a plugin using XSkull as: *"after a player's death [the plugin builds a head of the victim], an offline-mode UUID for them is added to usercache.json” works fine on a standalone server, only happens behind Velocity in online mode."*

**Impact:** the offline-UUID entry overwrites the correct name-to-UUID mapping the server itself wrote at login. Everything that resolves names through the user cache then returns the wrong UUID for that player: vanilla whitelist/ban/op resolution, `Bukkit.getOfflinePlayer(String)` in any plugin, and so on. Entries persist for a month, so the breakage outlives restarts.

### Root cause

All profile-identity decisions were based on `PlayerUUIDs.isOnlineMode()`, i.e. `Bukkit.getOnlineMode()`” which is `false` on *every* proxied backend, regardless of the proxy's own mode:

- `PlayerProfiles#sanitizeProfile` replaced the real UUID of a freshly fetched Mojang profile with the computed offline UUID before `MojangAPI#cacheProfile` inserted it into the server's user cache. This is the write path of the corruption; it triggers on any `Profileable.of(UUID)` lookup that has to fetch textures (user cache entries never contain textures, so effectively every first lookup).
- `MojangAPI#getMojangCachedProfileFromUsername` and `#usernamesToUUIDs` cached username lookups under the offline UUID the same way.
- `MojangAPI#getCachedProfileByUUID` translated the requested (real) UUID through `ONLINE_TO_OFFLINE` before querying the user cache, so lookups could miss the server's own real-UUID entries.

### The fix

A new `PlayerUUIDs#usesRealUUIDs()` determines which UUIDs the server actually deals in; all seven former `isOnlineMode()` decision sites now use it. Detection order:

1. **`xseries.profiles.proxyOnlineMode` system property**” manual override for setups the detection can't figure out.
2. **`Bukkit.getOnlineMode()`**” `true` means real UUIDs, as before.
3. **Paper's proxies config**” `GlobalConfiguration.get().proxies.isProxyOnlineMode()` (1.19+) or `PaperConfig.isProxyOnlineMode()` (1.12“1.18.2), via reflection. This is the exact signal Paper itself uses for its own name-based profile resolution, and it covers both Velocity and BungeeCord forwarding.
4. **Spigot behind BungeeCord** (`settings.bungeecord`): inferred from a currently online player” proxies compute offline UUIDs with the same formula as vanilla, so a player whose UUID differs from `getOfflineUUID(name)` proves an online-mode proxy. While nobody is online, the old assumption (offline) is used but *not* cached, so it re-evaluates on the next call.

The verdict is cached once definitive, since the mode can't change without a restart.

Also included:

- `getRealUUIDOfPlayer(String, UUID)` now trusts the given UUID on real-UUID servers, unless it happens to be the computed offline UUID. On proxied backends this removes one Mojang username-to-UUID request per profile.
- Bugfix in `usernamesToUUIDs`: the `UserCache_profilesByName.containsKey(name)` guard used the case-corrected name from Mojang's response, but the user cache keys that map by *lowercased* name” the guard never matched any username containing an uppercase letter, so profiles were re-cached unnecessarily on every call.

### Behavior matrix

| Server setup | Before | After |
|---|---|---|
| Standalone, online mode | real UUIDs | unchanged |
| Standalone, offline mode | offline UUIDs | unchanged |
| Paper behind online-mode proxy (Velocity/Bungee forwarding) | offline UUIDs written’ user cache corrupted | real UUIDs |
| Paper behind offline-mode proxy (declared in `proxies` config) | offline UUIDs | unchanged |
| Spigot behind online-mode BungeeCord | corrupted | real UUIDs (via player sample) |
| Spigot behind offline-mode BungeeCord | offline UUIDs | unchanged (player UUIDs match the offline formula) |

### Testing

Compiled with the full reactor (`mvn install -DskipTests`). The reflection targets were checked against current Paper sources (`GlobalConfiguration.Proxies#isProxyOnlineMode`). I have not live-tested every row of the matrix; the detection is written to fall back to the previous behavior whenever it can't produce a definitive answer, and standalone servers short-circuit at step 2 without any reflection.


CryptoMorin:
Absolutely a brilliant find and a lovely PR for it.

Now I understand the weird behavior I was seeing in proxy servers.
I'll have to take one final look later before merging the PR.

Thank you so much!

Krakenied:
Not sure if it's useful, but it's worth mentioning that from 1.21.5 there's a Paper method:
- https://jd.papermc.io/paper/26.2/org/bukkit/Bukkit.html#getServerConfig()
- https://jd.papermc.io/paper/26.2/io/papermc/paper/configuration/ServerConfiguration.html#isProxyOnlineMode()

which returns actual online mode state

image

